PR #1, opened 2023-07-10
Das/v02

mezarque:
This PR updates the package to be a little more sensibly structured, with colors, functions, and classes placed into different scripts.

This PR also introduces the `Palette` and `Gradient` classes, which provide methods to group colors and convert them between different data structures in a tidier way. These classes also have a variety of methods to display color palettes. Combining the class methods with additional built-in functions should make it easier for others to build their own custom color palettes and test them.

An explainer of how the package works can be found in the README.ipynb or README.md files.


README.md
```diff
@@ -37,303 +74,478 @@ display(apc.arcadia_Accent)
...
-When coloring heatmap images, it's often helpful to have perceptually uniform colormaps with multiple colors along a gradient.  
-Matplotlib provides several of these, such as `viridis`, `magma`, `cubehelix`, and others.  
 
-`arcadia_pycolor` includes two nearly perceptually uniform colormaps that have similar properties to `viridis` and `magma` but play nicer with the Arcadia color palette. They can be accessed via:
-- `arcadia:viridis`: similar to `viridis` but uses some Arcadia colors
-- `arcadia:magma`: similar to `magma` but looks better next to Arcadia colors
+---
+## 5. `matplotlib` features
+
+The `arcadia_pycolor` automatically integrates with `matplotlib`.
```
mertcelebi, line 341: How does this work exactly? Is there a way to to make this not automatic (ie implicit) but explicit?

README.md
```diff
@@ -2,33 +2,70 @@
 Tools for using the Arcadia palette in Python.  
 This package automatically generate color palettes and color maps for use with Matplotlib.
 
-# Installation
+---
```
mertcelebi, line 5: Why do we need both a markdown and a jupyter notebook README?

setup.py
```diff
@@ -1,15 +1,27 @@
...
 
 setup(
     name='arcadia_pycolor',
     url='https://github.com/Arcadia-Science/arcadia-pycolor',
     author='Dennis Sun',
     author_email='dennis.sun@arcadiascience.com',
     packages=['arcadia_pycolor'],
-    install_requires=['numpy', 'matplotlib', 'scipy'],
-    version='0.1',
+    install_requires=['numpy', 'matplotlib', 'scipy', 'seaborn', 'plotly'],
```
mertcelebi, line 14: How big are some of these libraries? It's a little odd that a color library has seemingly big plotting dependencies

arcadia_pycolor/classes.py
```diff
@@ -0,0 +1,301 @@
...
+        Randomly samples the color dictionary, returning the data in the desired format.
+        
+        Args:
+            number (int): number of random colors to return.
+            fmt (str): 'list', 'dict', 'nested_list', or 'tuple_list'
+        '''
+        if number > len(self):
+            raise Exception(f'Requested {number} colors but there are only {len(self)} in this palette.')
+        
+        if how == 'list':
```
mertcelebi, line 98: how or fmt?

arcadia_pycolor/classes.py
```diff
@@ -0,0 +1,301 @@
...
+            return random.choices(self.colors, number)
+        elif how == 'dict':
+            return {choice: self.dict[choice] for choice in random.choices(self.nicknames, number)}
+        elif how == 'nested_list':
```
mertcelebi, line 102: Just out of curiosity, what made you implement at this level of granularity? Are these pretty common in mpl? Seems so overkill to me

arcadia_pycolor/classes.py
```diff
@@ -0,0 +1,301 @@
...
+        '''
+        Uses display_palette to display the colors in the reversed palette.
+        '''
+        display_palette([{'name': self.name, 'length':  len(self.dict), 'cmap': self.mpl_ListedColormap_r}])
+        
+    def display_interactive(self):
+        '''
+        Uses display_palette_interactive to display an interactive hover-over version of the palette.
+        '''
+        display_palette_interactive({self.name: self.dict})
```
mertcelebi, line 179: nice

arcadia_pycolor/classes.py
```diff
@@ -0,0 +1,301 @@
...
+        if len(values) == 0:
+            fraction = 1 / (len(color_dict) - 1)
+            values_passed = [i * fraction for i in range(len(color_dict))]
+        # otherwise, checks to make sure the colors and values are of equivalent length
+        elif len(values) != len(color_dict):
+            raise ValueError('Number of values must be equal to number of color_dict entries.')
+        elif max(values) != 1 or min(values) != 0:
+            raise ValueError('Values must be bounded by and include 0 and 1.')
+        else:
+            values_passed = values
```
mertcelebi, line 210: What if values was passed as a string? Do you wanna detect that and turn it into an array?

mertcelebi (changes requested):

README.md
```diff
@@ -2,33 +2,70 @@
...
+apc.print_color(apc.aegean)
+```
+
+
+<span style="font-family: monospace"> #5088C5 <span style="color: #5088C5">██████</span> arcadia:aegean</span>
+
+
+### Palette Dictionaries
+
+The base color palettes (`arcadia_Core`, `arcadia_Neutral`, `arcadia_Accent`, etc.) are also stored as dictionaries.
```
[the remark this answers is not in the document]
mezarque: This makes sense, I'll remove `arcadia_` from most of these variable names.

The `"arcadia:"` prefix for the names of the colors and dictionaries should probably stay, as this colon prefix is analogous to what is used for colors in matplotlib such as `"tab:blue"` or `"xkcd:sky blue"` [see documentation](https://matplotlib.org/stable/tutorials/colors/colors.html).

Since some of Arcadia's color and palette names overlap with existing matplotlib names, the colon prefix prevents ambiguity and avoids overriding the matplotlib defaults.

README.md
mezarque: The markdown is generated from the Jupyter notebook. Unfortunately Github's default Markdown renderer doesn't appropriately format some of the outputs of the Jupyter notebook cells, so I've left both in there for people to see.

README.md
```diff
@@ -2,33 +2,70 @@
...
+
+
+```python
+apc.aegean
+```
+
+
+
+
+    {'arcadia:aegean': '#5088C5'}
```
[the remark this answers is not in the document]
mezarque: This is to prevent ambiguity with other built-in matplotlib colors and is meant to be analogous to the matplotlib [color naming convention](https://matplotlib.org/stable/tutorials/colors/colors.html).

arcadia_pycolor/classes.py
mezarque: I built these different structures based on my experience dealing with color format conversion between different Matplotlib plotting methods and also different Plotly plotting methods. For example, Matplotlib expects gradients as a list of tuples, whereas Plotly expects gradients as a list of lists. Categorical and continuous plotting methods in each software also expect colors in different ways. In some cases the plotting functions can be either require list or a dict depending on exactly what you're trying to do.

It's super annoying to deal with all of those conversions, so hopefully providing these reformatted structures as properties makes the conversions a little less tedious in the future.

arcadia_pycolor/classes.py
mezarque: I'm not sure when values would make sense to be string-formatted; it's meant to be a list of the values at each position along the gradient, so it should always have more than one value. Could you provide an example of how you imagine this occurring?

setup.py
mezarque: I can probably remove seaborn for now, since it's not being used yet. The original intent was to have a unified way of managing color that is compatible across different common Python plotting packages since the expected color formats can be different depending on the packages. 

I was also hoping to build in style sheets for Plotly eventually, so that plots are pre-configured to match the Arcadia default style on import.

Perhaps I could check to see whether some of these packages are installed and conditionally load relevant functions and methods depending on whether they are present.

README.md
```diff
@@ -37,303 +74,478 @@ display(apc.arcadia_Accent)
...
+print(apc.Palettes['arcadia:Accent'].dict)
+
+# Return colors as a list
+print(apc.Palettes['arcadia:Accent'].colors)
+
+# Return palette colors as a list of tuples
+print(apc.Palettes['arcadia:Accent'].tuple_list)
 ```
 
+    {'arcadia:aegean': '#5088C5', 'arcadia:amber': '#F28360', 'arcadia:seaweed': '#3B9886', 'arcadia:canary': '#F7B846', 'arcadia:aster': '#7A77AB', 'arcadia:rose': '#F898AE'}
```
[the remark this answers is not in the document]
mezarque: fixed!

README.md
```diff
@@ -37,303 +74,478 @@ display(apc.arcadia_Accent)
...
+Different plotting software might expect color information in different formats, including `list`s, `dict`s, `tuple`s, etc.  
+The `Palette` class allows for quick access to a variety of data structures.  
+
+
+```python
+# Return palette colors as a dictionary
+print(apc.Palettes['arcadia:Accent'].dict)
+
+# Return colors as a list
+print(apc.Palettes['arcadia:Accent'].colors)
```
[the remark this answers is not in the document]
mezarque: I felt like `colors` was maybe more intuitive, but `list` makes sense in the context of the other attributes.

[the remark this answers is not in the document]
mezarque: I'm going to leave `colors` in place but add an additional `list` property that returns the list of colors.

README.md
```diff
@@ -37,303 +74,478 @@ display(apc.arcadia_Accent)
...
+
+
+```python
+apc.Palettes['arcadia:Core'].display()
+apc.Palettes['arcadia:Neutral'].display()
+apc.Palettes['arcadia:AccentFull'].display()
+apc.Palettes['arcadia:LightFull'].display()
 ```
 
+    /Users/dennis/Code/arcadia-pycolor/arcadia_pycolor/functions.py:173: UserWarning: Tight layout not applied. The bottom and top margins cannot be made large enough to accommodate all axes decorations.
```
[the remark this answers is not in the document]
mezarque: This is fixed!

setup.py
mezarque: I're removed 'scipy' and 'seaborn', but I use matplotlib and plotly features within the package for visualization purposes.

README.md
mertcelebi: Yeah that makes sense. One note about this (and it's up to you what you go with):
* I am a little wary of "magical" things that libraries do that may impact other libraries' behavior. Because it's implicit and not explicit (ie done by the user), it may lead to confusing bugs or may overwrite things in unintentional ways. If I were to use this, I'd definitely prefer there to be utility functions for registrations, but for me to be the one calling those functions rather than the library calling it for me. Side effects are weird and lead to odd bugs.

But up to you, I don't mind it too too much

mertcelebi (approved):
Approving with a couple comments!

README.md
mezarque: This makes sense. At the end of the day, running one more line of code to get stuff registered isn't a huge deal, and it's better to provide the user the ability to determine where weird bugs might be coming from instead of untraceable "magic" happening in the background.

I've partitioned out the Matplotlib registration into a separate function, [`mpl_setup()`](https://github.com/Arcadia-Science/arcadia-pycolor/blob/fad77e62318e6f5f6247261be2f3aceb1f549bf6/arcadia_pycolor/colors.py#L288C1-L317C71), and updated the README to reflect this. I've set up the function to allow the user to specify more specifically what is being registered.

PR #2, opened 2023-07-13
added colorspacious as a dependency

mezarque:
Thanks to @MichaelReitman for spotting that `colorspacious` wasn't already called as an installation requirement!


MichaelReitman (approved):
Sweet!

PR #3, opened 2023-07-18
Das/v03

mezarque:
This PR adds the following features:
- built-in color vision deficiency testing for colors and images using [`colorspacious`](https://colorspacious.readthedocs.io/en/latest/tutorial.html#simulating-colorblindness)
- improved `arcadia:cividis` gradient
- strong monocolor gradients
- improved bicolor gradients


arcadia_pycolor/colors.py
```diff
@@ -140,13 +145,18 @@
 The colors have also been optimized to be nearly-perceptually uniform based on lightness.
 """
 
-viridis = {
+viridis_v1 = {
```
mertcelebi, line 148: what are the differences here?

arcadia_pycolor/functions.py
```diff
@@ -369,3 +373,81 @@ def plot_color_lightness(cmap_dict: dict, title = None,
         return fig
     
     plt.show()
+    
+    
+def make_cvd_dict(form: str, severity: int):
```
mertcelebi, line 378: awesome

arcadia_pycolor/functions.py
```diff
@@ -6,12 +6,13 @@
 from plotly.subplots import make_subplots
 from IPython.display import Markdown, display
 import colorsys
-from colorspacious import cspace_converter
+from colorspacious import cspace_converter, cspace_convert
```
mertcelebi, line 9: Are the functions in this file exposed to the end-user? Or there mostly for internal use? If the latter, `utils` maybe a better name

arcadia_pycolor/functions.py
```diff
@@ -369,3 +373,81 @@ def plot_color_lightness(cmap_dict: dict, title = None,
...
+        rgb_colors = [mpl.colors.to_rgb(i) for i in colors]
+    else:
+        rgb_colors = colors
+    
+    cvd_rgb_colors = np.clip(cspace_convert(rgb_colors, cvd_space, "sRGB1"), 0, 1)
+    cvd_hex_colors = [mpl.colors.to_hex(color) for color in cvd_rgb_colors]
+
+    return cvd_hex_colors
+
+def simulate_cvd_image(file: str, severity = 100, fig_width = 10, show = True):
```
mertcelebi, line 410: This seems more like for testing than not, but probably fine to have here

mertcelebi (changes requested):
A couple questions + suggestions

arcadia_pycolor/colors.py
mezarque: I added the v1 version to keep as a deprecated version - the new "viridis" has a purple-ish color towards the end of the gradient. I can probably remove the v1 version now.

arcadia_pycolor/functions.py
mezarque: The plan was to expose these to the end-user; the functions listed under `__all__` are made available under the package namespace when using `import arcadia_pycolor` or `import arcadia_pycolor as apc`.

mezarque:
Sorry for the long delay on getting to this! I went through and addressed the comments and also added some documentation + linting for the scripts.

mertcelebi (approved):
I am going to approve with a couple comments:
* Can you extend the README for setup instructions for people trying to improve on this locally? Like how to install the right packages etc?
* And make sure to proof-read the README?
* Is there a way to pin the underlying package versions?
* Can you add an MIT license?
* More discussions-y: for stuff like `simulate_cvd`, do you expect the average user to use those? or are there in the package for development reasons?

mezarque:
- added a "For Developers" section and a "Contributing" section to the README
- added a YML for the conda env used to develop the package
- pinned package versions to YML and `setup.py`
- added MIT license
- `simulte_cvd` is meant for general users to check whether their figures are colorblind-friendly, yes. Something all scientists generating figures should probably be doing, but I think Publishing is doing that currently. Hopefully this helps streamline / centralize that

PR #5, opened 2023-12-13
Remove font weight parameter

mezarque:
The font weight parameter in the mpl stylesheet can throw errors if a font with the exact weight value is not found on the system. This removes that parameter.


keithchev (approved):
ugh fonts. LGTM!

PR #8, opened 2024-05-17
Start of rebuilding the package

mezarque:
## PR checklist

# Overview
This PR begins the process of reworking this repository into a more robust Python package. As part of this, I'm going to refactor the core classes, starting with the Palette class and the new Color class. I'm temporarily adding a "_new" suffix to any modules that will eventually replace old modules and will gradually migrate as much necessary functionality from those old modules as possible into a new, more principled organization.

## Design Choices
Keith and I chatted a bit offline about some high- and low-level design choices for this package. We agreed that:
- Compartmentalizing different functions into separate modules would be ideal (e.g. `matplotlib`-related functions should live in their own module.)
- To transition to the new way of doing things, I'll mostly rebuild functionality from scratch, rather than modifying the existing repo.
- We might make exceptions to the above to streamline usability for special cases. For example, rather than storing the individual colors in a `colors` module and having users type `apc.colors.aegean`, we settled on having those colors live in the highest level of the namespace, e.g. as `apc.aegean`.
- We should avoid having too many dependencies.
- Keith will handle the packaging and PyPI distribution architecture and I'll focus on the core functionality.

## Changes
- Introduce an updated `Palette` class with reduced functionality (for now).
- Introduce a new `Color` class that allows users to see colors more easily, which are collected into `Palette` objects.
- Re-factor the existing `colors.py` collection of constants using `colors_new.py`; Gradients are yet to be implemented.
- Add a `display.py` module that will be used to collect all color display functions.
- No new dependencies.

## Testing
- You can see the functionality behaving as expected in the `usage_example.ipynb` notebook.

## Questions
- Currently, `Color` inherits from `str` so that it can be properly handled as a HEX code string by downstream libraries (e.g. `matplotlib`, `seaborn`, `plotly`). This works and has a better UX than having to call a method to get the HEX code, but I'm not sure if it's a good idea.
- Should we be using `pydantic` or something to have easier validation of input types? I tried implementing this, but the `Color` `str` inheritance made it tricky to figure out, so I abandoned it.
- Palettes are currently stored at the highest level of the namespace using `ALLCAPS` variable names. What's the preferred way to store them? Should they be in their own module? What should our naming convention be?
- What should the naming convention be for colors? I think the `apc.lowercase` UX is the simplest, but does it go against the expectation that constants are in `ALLCAPS`?
- What tests should we write for this package? I'm happy to do it, but it will take some time for me to get up to speed.

## Next
After this PR, I'll start working on:
- [ ] Implementing the `Gradient` class.
- [ ] Adding in methods to allow users to display `Gradient` colors.

Then, I'll put together:
- [ ] A module for styling `matplotlib` plots with style sheets and various helper functions.


mezarque:
Something fun about the way that `__repr()__` is now implemented is that you can see colors even when running an interactive Python session in the terminal.

Screenshot 2024-05-17 at 4 42 54 PM

keithchev (changes requested):
This looks like a great start! I love the swatches and it's so cool that they work basically everywhere.

I made an initial round of comments, mostly about implementation. For context, I'm trying to apply a pretty high bar to the code, since we're imagining that this will become a long-lived project and a dependency of many other projects at Arcadia.

My takes on the questions from the PR description:

> * Currently, `Color` inherits from `str` so that it can be properly handled as a HEX code string by downstream libraries (e.g. `matplotlib`, `seaborn`, `plotly`). This works and has a better UX than having to call a method to get the HEX code, but I'm not sure if it's a good idea.

I commented in detail on this in an inline comment, but TLDR, I don't think it's a good idea.

> * Should we be using `pydantic` or something to have easier validation of input types? I tried implementing this, but the `Color` `str` inheritance made it tricky to figure out, so I abandoned it.

For relatively simple validation (e.g. of a hex code), pydantic probably isn't worth it, imo. I would say we can make a note to revisit this later on, especially if we find that writing validation logic is becoming cumbersome. 

> * Palettes are currently stored at the highest level of the namespace using `ALLCAPS` variable names. What's the preferred way to store them? Should they be in their own module? What should our naming convention be?

I think they should be defined in their own module `palettes.py` and namespaced under `apc.palettes` so that they are clearly separated from the colors. And I think all-lowercase names are fine (see below).

> * What should the naming convention be for colors? I think the `apc.lowercase` UX is the simplest, but does it go against the expectation that constants are in `ALLCAPS`?

I think lowercase is okay, even though it breaks with convention since the `Color` instances are technically constants. This feels like a gray area to me because the `colors` module contains nothing but color definitions, so the colors (and palettes) "feel" closer to class attributes (where the "class" is actually a module) than constants. Also, practically, it means there's not a need to distinguish between constants and functions (this would be even more true if colors were namespaces under `apc.colors`, but I think lowercase is okay even if they're not).

> * What tests should we write for this package? I'm happy to do it, but it will take some time for me to get up to speed.

I would say that tests for input validation would probably be the most important, since those deal with user-provided values that could be weird. And then tests for conversion methods like hex-to-rgb, since those are so fundamental. But I think this package is simple enough that we can afford to wait on this for a little while until we have a stable-ish public API, and then we can go back add tests (also happy to do this and/or help!)

arcadia_pycolor/colors_new.py
```diff
@@ -0,0 +1,152 @@
...
+
+# Other named colors
+w = Color("w", "#FFFFFF")
+r = Color("r", "#FF0000")
+g = Color("g", "#00FF00")
+b = Color("b", "#0000FF")
+c = Color("c", "#00FFFF")
+m = Color("m", "#FF00FF")
+y = Color("y", "#FFFF00")
+k = Color("k", "#000000")
```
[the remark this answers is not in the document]
mezarque: This is fair; one problem is that Arcadia's color palette already has a `black` entry that is not `#000000`. So we'd need to rename that `black` to something else. But I think that was the main reason I used these abbreviations.

arcadia_pycolor/classes_new.py
```diff
@@ -0,0 +1,72 @@
+import matplotlib.colors as mcolors
+
+
+class Color(str):
```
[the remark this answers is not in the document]
mezarque: I think this makes sense to me overall from a software perspective, but it does make the user experience somewhat more annoying; needing to type `.hex_code` after every instance of a color usage feels like it will get wordy over time. For example, between `apc.tangerine` and `apc.tangerine.hex_code`, the latter feels pretty ugly/ unwieldy. I've updated this so that the `str` inheritance is removed, but given that direct access to colors will likely be one of the most-used features, I wonder if we can figure out some way to get the `apc.tangerine` syntax to work properly without inheriting from `str`.

arcadia_pycolor/colors_new.py
```diff
@@ -0,0 +1,152 @@
...
+
+# Core palette
+CORE = Palette(
```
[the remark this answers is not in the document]
mezarque: How should I deal with importing the colors? If I don't use * import, will I have to list all of the color names?

arcadia_pycolor/colors_new.py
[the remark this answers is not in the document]
mezarque: I'm going to rename the existing `black` to `pitch` and go with the fully spelled-out names.

arcadia_pycolor/__init__.py
```diff
@@ -1,5 +1,5 @@
-__all__ = ["classes", "functions", "colors"]
+__all__ = ["classes_new", "colors_new", "display"]
 
-from .classes import *
-from .colors import *
-from .functions import *
+from .classes_new import *
```
mertcelebi, line 3: Why do we have to name these like that? I'd prefer to name these as `classes` and name the old ones as `classes_old`. 

If it's to avoid weird version issues with installed instances of `arcadia-pycolor`. It may be a good idea to merge these changes to a feature branch other than main. And when we're ready to merge to main, we can go from feature branch to main.

pyproject.toml
```diff
@@ -23,6 +23,7 @@ extend-include = ["*.ipynb"]
 # ignore useless expressions used for printing examples
 # ignore module level import not at top of file
 "README.ipynb" = ["B018", "E402"]
+"usage_example.ipynb" = ["B018", "E402"]
```
mertcelebi, line 26: See this thread: https://arcadiascience.slack.com/archives/C04AKK76CDN/p1716233305279529

I think we should update the ruff version of the repo and not explicitly ignore these "rules"

keithchev:
Two more comments I forgot to include:
- it looks like the env files are out of date. `dev.yml` looks like it might have been inherited from the python template, since it has nothing specific to this repo, and `apc.yml` uses python 3.9, but the type hints require python 3.12. I figured I should mention this but I don't know if it's worth dealing with this now, since the deps will have to be moved into `pyproject.toml` anyway to make the package pip-installable. 
- since we're breaking the transition between "old" and "revamped" versions of the package into multiple PRs, it would be a good practice to merge PRs to a branch other than `main`, so no one uses the package in its current in-between state (this isn't a huge deal since the "old" package is still usable; it's more of a best practice that errs on the side of caution).

arcadia_pycolor/classes_new.py
```diff
@@ -0,0 +1,61 @@
...
+        self.name = name
+        self.colors = colors
+
+    def from_dict(self, colors: dict[str, str]):
+        self.colors = [Color(name, hex_code) for name, hex_code in colors.items()]
+
+    def __repr__(self):
+        from arcadia_pycolor.display import swatch  # imported here to avoid circular import
+
+        longest_name = max(len(color.name) for color in self.colors.values())
```
mertcelebi, line 53: Given the definition of colors is a "dict or a list", this should handle different types

arcadia_pycolor/classes_new.py
```diff
@@ -0,0 +1,61 @@
...
+            raise ValueError(f"Invalid HEX code: {hex_code}")
+
+        self.name = name
+        self.hex_code = hex_code
+
+    @property
+    def to_rgb(self):
+        return [int(c * 255) for c in mcolors.to_rgb(self.hex_code)]
+
+    def __repr__(self):
```
mertcelebi, line 25: I am by no means a Python expert, so excuse my naive question but this seems a little unconventional to me. I would have expected `__repr__` to be a debug friendly version of the class so it can be re-instantiated, and `__str__` to be more like this. Am I wrong?

arcadia_pycolor/classes_new.py
```diff
@@ -0,0 +1,61 @@
...
+            colors (dict): a dictionary where the key is the color's name as a string
+                and the value is the HEX code of the color as a string
+            OR
+            colors (list): a list of Color objects
+        """
+        self.name = name
+        self.colors = colors
+
+    def from_dict(self, colors: dict[str, str]):
+        self.colors = [Color(name, hex_code) for name, hex_code in colors.items()]
```
mertcelebi, line 48: The below comment applies to this as well

mertcelebi (commented):
Keith already gave a pretty thorough review, so leaving some comments/questions. 

One note I have is around modules. I think having all the colors be namespaced under `apc.colors` and palettes under `apc.palettes`. Yes, it's longer to type, but feels more natural/logical to me.

arcadia_pycolor/classes_new.py
mertcelebi: While at it, could you also explain what this means?

"Currently, Color inherits from str so that it can be properly handled as a HEX code string by downstream libraries (e.g. matplotlib, seaborn, plotly). This works and has a better UX than having to call a method to get the HEX code, but I'm not sure if it's a good idea."

Does it mean, when `Color` inherited from `str`, you could pass it to these downstream tools directly with `apc.blah` and now you need to do `apc.blah.hex_code`?

keithchev: > I am by no means a Python expert, so excuse my naive question but this seems a little unconventional to me. I would have expected `__repr__` to be a debug friendly version of the class so it can be re-instantiated, and `__str__` to be more like this. Am I wrong?

FWIW my take on this is that returning a color swatch adheres to the *spirit* of the `__repr__` method, which I'm interpreting to be that it should return a succinct representation of the object that is useful for debugging. In this special case of color, the argument would be that the swatch is more meaningful and interpretable than something like `Color(hex_code="#fff")` for most users. There is some precedent for this with e.g. pandas dataframes, whose `__repr__` method returns a truncated and pretty-printed dataframe rather than `pd.DataFrame(data=...)`.

keithchev: There are also two pragmatic considerations: 
- `__repr__` is what is called by jupyter notebooks to evaluate expressions, so I think it's the only way to show the swatch when the user types a color like `apc.tangerine` in a notebook cell.
- if we want `str(color)` to return the hex code, we have to reserve `__str__` for returning the hex code itself.

keithchev:
@mezarque I chatted with Mert briefly about the subclassing-str thing, and we had an idea that we think would be user-friendly and also avoid the weirdness of a `Color` class that subclasses `str`. The idea is to avoid the need to define a `Color` class entirely by defining utility functions for color operations like `lighten` and `darken` etc. We thought this would still be ergonomic/intuitive since the user would write `lighten(apc.tangerine)` rather than `apc.tangerine.lighten()`. And of course we could define the colors themselves as instances of a `HexCode` class that inherits from string (as we discussed yesterday) so `apc.tangerine` could be passed directly to matplotlib plots etc.

arcadia_pycolor/__init__.py
mezarque: I agree that we should be merging into a feature branch rather than main; I'm going to address PR comments here until we're satisfied, then close this PR and open a fresh one to a feature branch.

arcadia_pycolor/classes_new.py
mezarque: We decided to just have `self.colors` be a list, so I've refactored for that constraint.

arcadia_pycolor/classes_new.py
mezarque: Ditto all of what Keith said above! In general I am trying to aim for the best usability, and am willing to advocate for some sacrifices in the purity of the architecture if it means we can make the package really clear and simple for downstream users.

mezarque:
Thanks for all your feedback! I'm learning a lot.
I made a variety of changes as suggested:

- There is now a `palettes` module.
- The `Color` class is now `HexCode` and inherits from `str`, replacing its `__repr__` so we can see color swatches.
- Renamed old code to have `_old` suffix.
- The `Palette` class now by default accepts a list of `HexCode` objects. To instantiate from a `dict` of `str`, there's now the `from_dict()` classmethod.

Some remaining points of discussion:
- I haven't removed the `to_rgb` method from `HexCode`. It doesn't feel that unnatural to me to keep it there.
- I have kept all the colors within the high-level namespace, e.g. they're accessed as `apc.aegean` rather than `apc.colors.aegean`. I am pretty resolute in wanting to keep that structure, because I expect users will be accessing colors a great deal.

If this sounds OK, I'm going to close this PR and open a new one on a feature branch so we avoid breaking everything in main.

keithchev (approved):
Looks good! Thanks for your patience with my engineering-y-ness 😄 

Agree that `to_rgb` feels okay on `HexCode`. But I feel we should draw the line at methods like `lighten` and `darken` since (as we discussed) that opens the door to a large number of color-specific methods and potential confusion.

Aside: I noticed the swatches for very light colors need some work, since they're invisible if the background is white (e.g. a jupyter notebook in light mode). And I guess the opposite will be true in dark mode. But something for a future PR.

> If this sounds OK, I'm going to close this PR and open a new one on a feature branch so we avoid breaking everything in main.

I think there should be a way to avoid doing this (which would be preferable to preserve the history): create a new branch at `main`, push it to the remote, then change the "base" of this PR to that new branch (by clicking the "edit" button next to the PR title).

PR #9, opened 2024-05-22
New implementation of gradients

mezarque:
## PR checklist

## Overview

This PR introduces a reworked `Gradient` subclass, which inherits from the `Palette` class. The main difference is that the `Gradient` subclass also has a list of positions associated with each color, on a scale from 0 to 1, which describe the gradient.

## Changes
- Fixed some color names (e.g. `b` -> `black`)
- Added `Gradient` subclass.
- Added `gradient_swatch` display utility, which creates a special swatch that shows a range of gradient steps as part of the `__repr__` for the `Gradient` class.
- Added `plot_gradient_lightness` display utility, which converts a gradient to HSL colorspace and plots lightness; useful when creating new gradients. I expect this to be used almost exclusively by myself and Audrey, but it's something she really wanted access to.
- Added an `mpl` module with `to_cmap` and `to_linear_cmap` methods that create `matplotlib`-relevant objects that are useful for interpolating gradients (and which will be used to load colors, gradients, and palettes into `matplotlib` in the future).
- Define the same gradients that were available in the previous version in the `gradients` module. These may change again in the future, as Audrey is looking into what the right options are, but I wanted to put these in for now.

## Testing
I confirmed that importing works and functionality is behaving as expected, as recorded in `usage_example.ipynb`.

## Questions
- In the `gradient_swatch` method, I import `HexCode` locally to avoid a circular dependency error. The `if TYPE_CHECKING` statement alone results in an error because `HexCode` is undefined. How do I resolve this? Should the first parts of the `gradient_swatch` method live somewhere else?

## Next
Before adding in functionality for `matplotlib`, which I've already figured out, I wonder if it would be good to put some testing in place. I know Keith can do this, but I would also like to learn some testing basics. What are some good tests to write for the package as it is now?


arcadia_pycolor/display.py
```diff
@@ -34,6 +41,132 @@ def swatch(color: "HexCode", width: int = 2, min_name_width: int = None):
...
+    # Get the color for each step in the gradient
+    colors = [HexCode(i, cmap(i / steps)) for i in range(steps)]
+
+    swatches = [colorize(" ", bg_color=c) for c in colors]
+
+    return "".join(swatches)
+
+
+def plot_gradient_lightness(
+    gradients: Union["Gradient", str, list[Union["Gradient", str]]],
```
mezarque, line 67: I realize this is kind of a mess, but I don't want to force the user to unintuitively wrap a single gradient entry in a list in order for this function to work. What's the best way to do that?

arcadia_pycolor/display.py
```diff
@@ -34,6 +41,132 @@ def swatch(color: "HexCode", width: int = 2, min_name_width: int = None):
     return output
 
 
+def gradient_swatch(gradient: "Gradient", steps=21):
+    from arcadia_pycolor.classes import HexCode
```
mezarque, line 45: Is this fine? I can't import at the top because of a circular dependency error, and the `if TYPE_CHECKING` alone doesn't actually allow `HexCode` to be in the namespace, which results in an undefined error.

keithchev: Yeah, this should be avoided. I think in this case the circular import is a sign that things could be reorganized. There are two options that I think could work:
1) move `gradient_swatch` to a method of `Gradient` (it could be a static method if needed). For consistency I'd also move `to_linear_cmap`.
2) define `colorize` to accept a hex code as a `str` (so you don't need to import `HexCode` here). Since it seems like `colorize` is an internal utility and not part of the public API, this feels fine to me.

I think either would be fine but I'd vote for option (1) since it helps keep things a bit more organized, especially as it looks like `gradient_swatch` is intended for use in `Gradient.__repr__`, in which case it is properly a private instance method of `Gradient`.

arcadia_pycolor/display.py
```diff
@@ -34,6 +41,132 @@ def swatch(color: "HexCode", width: int = 2, min_name_width: int = None):
...
+        horizontal_spacing (float): the spacing between lines
+        steps (int): the number of steps along the gradient to generate
+        figsize (tuple): the width, height tuple of the figure size
+        cmap_type (str): 'linear' if you want the label for the cmap to be at the end;
+            anything else puts the label in the middle.
+        tickrotation (int): rotation of the label for each cmap
+        markersize (int): the size of the points that make up the gradient color line
+        return_fig (bool): whether or not to return the figure as an object
+    """
+    from arcadia_pycolor.classes import Gradient
```
keithchev, line 98: This nested import should also ideally be avoided. One way would be to move the plotting function out of this module and into something like `plotting`, which would separate it from `colorize` (which feels like a utility).

keithchev (changes requested):
Looks good! TLDR the nested import statements should be avoided if possible. I made some suggestions about this in inline comments.

One important thing: it looks like the `apc.utils` module is missing from the PR.

A few minor things:
- the `__all__` in the top-level `__init__.py` needs to be updated.
- consider renaming `swatch` to `color_swatch` for symmetry with `gradient_swatch` and adding a verb to the name like `make_color_swatch` to clarify that it's a function and not a variable.
- there are some missing type hints, particularly for return types. This is nbd for now, imo, as we can always go back and add them later.

Ideas for tests:
- tests for the low-level utility functions like `colorize` and `swatch`. Make sure they behave as expected for some hard-coded colors and don't work for some common kinds of invalid input (`None`, empty string, a string like `"red"`, an invalid hex code, etc). 
- tests for the core primitives (`HexCode`, `Palette`, `Gradient`). Instantiate them with hard-coded valid and invalid input both directly and using `from_dict`. Check that `__repr__` functions return the expected output, since that's an important part of the user-facing functionality. Check that two palettes can be added together, check that gradient-to-cmap conversion works as expected, etc.

arcadia_pycolor/__init__.py
```diff
@@ -3,4 +3,6 @@
 from .classes import *
 from .colors import *
 from .display import *
+from .gradients import *
+from .mpl import *
```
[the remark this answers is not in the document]
mezarque: Makes sense!

arcadia_pycolor/display.py
```diff
@@ -34,6 +41,132 @@ def swatch(color: "HexCode", width: int = 2, min_name_width: int = None):
...
+    ax.xaxis.set_major_locator(ticker)
+    ax.xaxis.set_tick_params(rotation=tickrotation)
+    ax.set_xticklabels(labels=grad_names)
+    ax.set_ylabel("Lightness $L^*$", fontsize=12)
+
+    if title is not None:
+        ax.set_xlabel(title, fontsize=14)
+
+    if return_fig:
+        return fig
```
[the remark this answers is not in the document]
mezarque: This is made available so that a user can save, annotate, or otherwise modify the figure if they'd like to. For example, `plt.savefig()` must be called before `plt.show()`. Probably won't be used by most folks, but it's good to have the option (and saves them from having to reproduce this plotting code just to be able to handle the figure).

arcadia_pycolor/display.py
```diff
@@ -34,6 +41,132 @@ def swatch(color: "HexCode", width: int = 2, min_name_width: int = None):
...
+
+        grad_names.append(name)
+
+        # Get RGB values for colormap and convert the colormap in
+        # CAM02-UCS colorspace.  lab[0, :, 0] is the lightness.
+        if name in mpl.colormaps:
+            rgb = mpl.colormaps[name](x)[np.newaxis, :, :3]
+        else:
+            rgb = cmap(x)[np.newaxis, :, :3]
+        lab = cspace_converter("sRGB1", "CAM02-UCS")(rgb)
```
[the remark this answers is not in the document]
mezarque: This is great; thanks for untangling this! I was modifying some old code and it was simpler just to add more conditionals, but certainly not elegant.

arcadia_pycolor/display.py
mezarque: For some reason my Python 3.12 is throwing invalid operator errors with `|`, so I just used `Union` to get around that. Not sure what's going on, but we can revisit once I've figured out what's going on with my machine/ Python kernel. `Union` might also enable us to support older Python versions that don't support the `|` operator.

arcadia_pycolor/display.py
```diff
@@ -34,6 +41,132 @@ def swatch(color: "HexCode", width: int = 2, min_name_width: int = None):
...
+
+    # Get the color for each step in the gradient
+    colors = [HexCode(i, cmap(i / steps)) for i in range(steps)]
+
+    swatches = [colorize(" ", bg_color=c) for c in colors]
+
+    return "".join(swatches)
+
+
+def plot_gradient_lightness(
```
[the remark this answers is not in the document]
mezarque: In this case, the plotting works both for a singular and a plural number of gradients, so I think the non-plural form makes more sense.

arcadia_pycolor/display.py
mezarque: Should I also move `to_cmap()` to `Palette` then? My only concern there is that this mixes the `matplotlib`-specific functionality into the class. Given that we'd eventually like this to work with `plotly`, for example, a generic `to_cmap` method of `Palette` or `Gradient` would probably be confusing. We could get around this by having methods like `to_mpl_linear_cmap()`, but that feels kind of verbose.

arcadia_pycolor/gradients.py
```diff
@@ -0,0 +1,136 @@
...
+)
+
+seaweedtangerine = Gradient(
+    "seaweedtangerine",
+    [colors.seaweed, colors.paper, colors.tangerine],
+    [0, 0.5, 1],
+)
+
+# Strong Bicolor Gradients
+poppies = Gradient(
```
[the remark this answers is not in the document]
mezarque: I wonder about the ergonomics of writing out these names actually; maybe an abbreviation scheme like `matplotlib` uses would be reasonable?

e.g. see [here](https://matplotlib.org/stable/users/explain/colors/colormaps.html)

arcadia_pycolor/display.py
mezarque: I ended up moving both `swatch()` and `gradient_swatch()` into class methods of `Color` and `Gradient` respectively. They're both called via `swatch()` now, because imo it would feel silly to call `.gradient_swatch()` on something you already know to be a gradient.

arcadia_pycolor/display.py
keithchev: I meannn, I kinda like the `to_mpl_linear_cmap` idea. But because `to_cmap` converts to a matplotlib-specific object, I think it would also make sense to leave it in `mpl`.

arcadia_pycolor/display.py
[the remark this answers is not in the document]
keithchev: Mmm I see. Fair enough!

arcadia_pycolor/gradients.py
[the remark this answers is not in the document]
keithchev: Yeah, I think the `RdBuGn`-style abbreviations could work, though otoh there are some pretty esoteric color names so the abbreviations could be too opaque. 

Separately, the names like `poppies` would be less ambiguous if the gradients were namespaced under `apc.gradients` so that there could be no mistaking `pac.gradients.poppies` for a color.

arcadia_pycolor/display.py
keithchev: Nice. Agreed; if they're class methods, the name `swatch` alone is fine.

arcadia_pycolor/gradients.py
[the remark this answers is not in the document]
mezarque: I've got them namespaced under `gradients` now so I'm going to revert to the flower names for now; probably the gradients will change soon anyway once Audrey gets back.

arcadia_pycolor/display.py
mezarque: I might move the `to_mpl_linear_cmap()` direction in the future, but let's just keep it in the `mpl` module for now; I think we can revisit once I've added the various utilities that will be included there and we can get a better sense of whether they should be separated out.

mezarque:
I'm going to try to implement testing in another PR, just to keep things tidier.

arcadia_pycolor/utils.py
```diff
@@ -0,0 +1,8 @@
+import numpy as np
+
+
+def distribute_values(objects, min_val=0, max_val=1):
```
keithchev, line 4: Could this function accept a length `num_points` instead of a list `objects`? This would be simpler (and easier to type-hint), since the function does not depend on the contents of the list but just its length.

keithchev (commented):
Just two more comments (sorry!)

arcadia_pycolor/utils.py
mezarque: can do!

arcadia_pycolor/__init__.py
```diff
@@ -1,6 +1,13 @@
-__all__ = ["classes", "colors", "display", "palettes"]
+__all__ = [
+    "classes",
+    "colors",
+    "gradients",
+    "mpl",
+    "palettes",
+    "plot",
+]
```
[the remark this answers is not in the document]
mezarque: One wonders if Tiktaalik could have foreseen the wonders of its descendants' lives upon the land...

Thanks for figuring out how to do this, it's implemented now!

arcadia_pycolor/classes.py
```diff
@@ -48,12 +80,75 @@ def from_dict(cls, name: str, colors: dict[str, str]):
...
+        return "\n".join(
+            [self.swatch()]
+            + [
+                f"{color.swatch(min_name_width=longest_name)} {value}"
+                for color, value in zip(self.colors, self.values)
+            ]
+        )
+
+
+def _get_longest_name(palette: Palette) -> int:
```
mertcelebi, line 152: Is there a reason this is outside the `Palette` class?

mertcelebi: Or maybe part of a util file?

mertcelebi (commented):
This looks good to me. Made a couple small comments. I'll let Keith approve when he does a final pass.

arcadia_pycolor/classes.py
mezarque: Inside Palette class feels tidier to me; I've moved it!

mezarque:
Thanks for all your feedback! I moved `_get_longest_name()` and made associated changes. Merging shortly!

PR #10, opened 2024-05-24
Use poetry for dependency management and build tooling

keithchev:
## Overview
This PR introduces poetry for dependency management and build tooling. The purpose of this is to allow us to publish the `arcadia-pycolor` package to PyPI so that it is pip-installable using `pip install arcadia-pycolor`.

This involves the following changes:

- update `pyproject.toml` according to the poetry docs and remove `setup.py`.
- move dependencies from `setup.py` and the conda env files into the appropriate sections of `pyproject.toml` and remove the env files.
- specify a minimum rather than an exact version for all dependencies, as the package needs to support a range of python and matplotlib versions. For now, I picked a minimum of python 3.9 (released in Oct 2020) and matplotlib 3.7 (released in Feb 2023).
- add a poetry lockfile for consistent dev envs (this won't be used when end users pip-install the package).
- add an env file to store API tokens for the PyPI servers.
- add makefile commands to build and publish the package.
- bump the version of the package (for now, I bumped from `0.3.1` to `0.4.0`).

## Drive-by changes
- remove the old README and the notebook `README.ipynb` from which it was derived. This was for a few reasons: I wanted to update the README directly with instructions about developing and publishing the package; the notebook included dependencies that are not used by the package itself; it looks like the notebook will be replaced by the new `usage_examples.ipynb` notebook. 
- add caching and a step to run the tests in the existing CI workflow.

## Testing
All of this was done manually.
- I tested that the package can be built and published to the PyPI test server.
- I tested that the package can be installed by `pip` from a local wheel and from the test server.
- I tested that the pip-installed package can be imported without errors and the color swatches work.
- I checked that the `mplstyles/` directory is included in the builds.


.env.copy
```diff
@@ -0,0 +1,5 @@
+# API token for the PyPI prod server.
+POETRY_PYPI_TOKEN_PYPI=
```
mezarque, line 2: Just curious, will this eventually need to be made secret?

pyproject.toml
```diff
@@ -1,52 +1,69 @@
...
+
+[[tool.poetry.source]]
+name = "pypi_test"
+# Note: poetry only support the "legacy" API.
+url = "https://test.pypi.org/legacy/"
+
+[tool.poetry.dependencies]
+colorspacious = ">=1.1.2"
+ipykernel = ">=6.25.2"
+matplotlib = ">=3.7.0"
```
mezarque, line 34: I double checked the [info I found](https://stackoverflow.com/questions/35668219/how-to-set-up-a-custom-font-with-custom-path-to-matplotlib-global-font) about the `matplotlib` fonts API and it seems like it was last changed at 3.2.0, so I think we shouldn't have to worry about that API change for this version range at least.

pyproject.toml
```diff
@@ -1,52 +1,69 @@
...
+]
 readme = "README.md"
-requires-python = ">=3.9"
-license = { file = "LICENSE" }
+license = "MIT"
+packages = [
+    { include = "arcadia_pycolor" },
+]
+include = ["mplstyles/*"]
+homepage = "https://github.com/Arcadia-Science/arcadia-pycolor"
```
mezarque, line 18: It would be great to eventually have a [`mkdocstrings`](https://mkdocstrings.github.io/)-based documentation homepage page. I previously made one for [`glial-origins`](https://arcadia-science.github.io/glial-origins/BioFile_tutorial/) which worked pretty nicely. 

The great thing about `mkdocstrings` is that the webpages are built by reading the package docstrings, so there's a single source of truth for the API that's very proximal to the code.

pyproject.toml
```diff
@@ -1,52 +1,69 @@
...
+requires = ["poetry-core>=1.0.0"]
+build-backend = "poetry.core.masonry.api"
+
+[tool.poetry]
 name = "arcadia-pycolor"
+version = "0.4.0"
 description = "A Python package to distribute Arcadia's color and style guidelines for figures."
-version = "0.0.0"
+authors = [
+    "Dennis Sun <dennis.sun@arcadiascience.com>",
```
mezarque, line 10: 😇 

```suggestion
    "Keith Cheveralls <keith.cheveralls@arcadiascience.com>",
    "Dennis Sun <dennis.sun@arcadiascience.com>",
```

mezarque (commented):
As far as I understand this, it looks good to me. I don't understand the vagaries of Python packaging, so I'll let Mert weigh in more. Otherwise, I added some thoughts for the future throughout.

Some adjacent questions and thoughts not related to a specific file:

- Should I be using `poetry` to manage my dev environment dependencies instead of `conda`? Could someone point me to their favorite resources for making that transition/ what the caveats are?
- I actually intended `usage_examples.ipynb` to just be a temporary document I'm using to illustrate the API in detail (and test its functionality in Jupyter notebooks), with the hope that actual demonstration / walkthrough of the API for a user will live somewhere else, such as in the README. Not sure what the best place for that might be, but a `mkdocstrings` Markdown page might be the easiest. I'd argue that we should have at least a small handful of usage examples in the README in addition to the installation instructions. But in any case, I'm fine with deleting all these files for now to declutter the repo.

.env.copy
mezarque: Also, did we ever figure out if we need a license to distribute a package using PyPI as a corporation?

keithchev: Yeah the tokens are secret and the env files are a way of keeping it that way. I didn't add this to the readme yet, but `.env.copy` is just a template. The idea is for individuals to copy `.env.copy` to `.env` in their local repo, and then they can enter their own API keys into `.env` (which is in `.gitignore` so it can't be committed). 

Not sure about the license q but maybe Mert knows?

pyproject.toml
keithchev: Yes, agreed!

.github/workflows/lint.yml
```diff
@@ -22,10 +23,11 @@ jobs:
       - name: Install test dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install pre-commit==3.5.0 ruff==0.4.4
+          pip install poetry==1.8.3
```
mertcelebi, line 26: I'd add caching to these dependencies a la the Compass repo

.env.copy
```diff
@@ -0,0 +1,5 @@
+# API token for the PyPI prod server.
```
mertcelebi, line 1: I'd update the dev instructions with this + how to use this

Makefile
```diff
@@ -10,3 +12,27 @@ format:
...
+clean:
+	rm -rf dist
+
+.PHONY: build
+build: clean
+	poetry build
+
+# Note: `poetry` does not appear to read the `POETRY_PYPI_TOKEN_<NAME>` environment variable,
+# so we need to pass it explicitly in the `publish` command.
+.PHONY: test-publish
```
mertcelebi, line 26: Should we also have a public GitHub action for specific deploys? Maybe to a prod branch or smth? 

Probably not needed for this version, but potentially could be added as a todo

mertcelebi (approved):

keithchev:
> Should I be using poetry to manage my dev environment dependencies instead of conda? Could someone point me to their favorite resources for making that transition/ what the caveats are?

Do you mean for apc or in general? For apc, I added instructions in the readme that should be easy to follow (I hope). For most projects, I think conda is fine. IMO poetry is only worth using for projects that are python packages that we need to build and release, as it provides build tooling (the `poetry build` and `poetry publish` commands) that makes this easier. Its strict dependency management (the lockfiles) is also nice, but conda env files come close enough. Plus many projects have dependencies that are only on conda-forge or bioconda, and poetry won't work for those.

> I actually intended usage_examples.ipynb to just be a temporary document I'm using to illustrate the API in detail (and test its functionality in Jupyter notebooks), with the hope that actual demonstration / walkthrough of the API for a user will live somewhere else, such as in the README. Not sure what the best place for that might be, but a mkdocstrings Markdown page might be the easiest. I'd argue that we should have at least a small handful of usage examples in the README in addition to the installation instructions. But in any case, I'm fine with deleting all these files for now to declutter the repo.

Agree re the mkdocs solution and having examples in the README. But we should find a way to avoid making `README.md` an artifact that has to be generated from a notebook, just because this is a bit messy and hard to maintain.

Makefile
keithchev: Yeah, I figured automation could come later, if needed. I'll open an issue for this.

PR #11, opened 2024-05-24
Adding in tests

mezarque:
## PR checklist

## Overview

This PR adds a variety of tests to the repo to ensure robustness. I tried to write tests that ensure that expected functionality behaves properly, as well as a smaller number of tests to handle invalid inputs. As part of this, I also reworked certain functionality that was bugged, which is a great demonstration of the importance of tests!

## Changes
- Don't rely on `mcolors.is_color_like()`, which returns `True` even for named color strings, e.g. `white` or `sky blue`. Instead, directly validate whether something is a HEX code using regex via the new `_is_hex_code()` function.
- Use `longest_name_length` instead of `longest_name` where appropriate, since it's not the name itself that is returned, but instead the number of characters in the name.
- More input validation for `Gradient` class.

## Testing
All the tests I've written are passing right now. I feel like there's more input validation that could be added to a variety of methods. It would be great to have a second opinion for parts of the API that might not behave as a user would expect and to write tests/ input validation logic to handle those.

## Questions
- How should we test plotting functions, which live in `plot.py`?
- How do we get GitHub Actions to run the tests? I think I can try to reverse engineer by looking at how it's implemented in `ProteinCartography` but my brain is too melted to do this today.

## Next
Probably we'll want to add more tests and evaluate code coverage (would love Keith's help on this!), but I've run out of steam on writing tests. After this, I'll share a somewhat larger PR that implements a whole slew of utilities for `matplotlib`, including handling fonts, registering colors, updating the stylesheet, etc.


arcadia_pycolor/tests/test_hexcode.py
```diff
@@ -0,0 +1,138 @@
...
+    assert _is_hex_code(hex_string) == is_hex_code
+
+
+@pytest.mark.parametrize(
+    "name, hex_code",
+    [
+        ("white", "#FFFFFF"),
+        ("white", "#ffffff"),
+        ("white", "#FFF"),
+        ("white", "#fff"),
```
keithchev, line 33: I'd add a few more here like "#000" and a "normal" hex code like "#f0a2c3" here just to be safe.

arcadia_pycolor/tests/test_hexcode.py
```diff
@@ -0,0 +1,138 @@
...
+    [
+        ("white", "#ZZZZZZ"),
+        ("white", "#ZZZ"),
+        ("white", "#FF"),
+        ("white", "apples"),
+        ("white", "123456"),
+        ("white", "ffffff"),
+        ("white", "white"),
+    ],
+)
```
keithchev, line 51: It might be worth defining constants `VALID_HEX_CODES` and `INVALID_HEX_CODES` and using them both here and above in `test_is_hex_code`, both for consistency and to avoid the repetition.

arcadia_pycolor/tests/test_hexcode.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+@pytest.mark.parametrize(
+    "name, hex_code, rgb",
+    [
+        ("white", "#FFFFFF", [255, 255, 255]),
+        ("aegean", "#5088C5", [80, 136, 197]),
+    ],
+)
+def test_hexcode_to_rgb(name, hex_code, rgb):
```
keithchev, line 126: Nit: It seems unlikely in this context, but one can imagine weird rounding errors that only happen near but not at 0 or 255. So adding some colors like "#FFFFFE" or "#001" could be good.

arcadia_pycolor/tests/test_import.py
keithchev, line 1: Thoughts on this module:
- the import statements should be moved out of the tests to the top of the module.
- I think the two tests are redundant, because `import arcadia_pycolor` is the same with or without the `as apc` alias.
- I would use a for-loop to avoid both the repeated `assert hasattr` lines.
- I'm not sure tests like these necessary. I *think* it's a bit unconventional to test imports like this since these should be defined by `__all__`, and they're arguably just testing python's import machinery.

arcadia_pycolor/tests/test_mpl_conversions.py
```diff
@@ -0,0 +1,71 @@
...
+    "name, input, hex_codes",
+    [
+        (
+            "my_palette",
+            [HexCode("white", "#FFFFFF"), HexCode("black", "#000000")],
+            [HexCode("white", "#FFFFFF"), HexCode("black", "#000000")],
+        )
+    ],
+)
+def test_palette_to_cmap(name, input, hex_codes):
```
keithchev, line 17: `input` and `hex_codes` parameters can be condensed into one variable, right? Looks like they should always be the same.

arcadia_pycolor/tests/test_mpl_conversions.py
```diff
@@ -0,0 +1,71 @@
...
+            [0, 1],
+        ),
+        (
+            "my_gradient",
+            [HexCode("white", "#FFFFFF"), HexCode("black", "#000000")],
+            None,
+        ),
+    ],
+)
+def test_gradient_to_linear_cmap(name, colors, values):
```
keithchev, line 49: On the one hand, this test could benefit from more test cases. On the other hand, since `gradient_to_linear_map` just wraps `LinearSegmentedCoolormap.from_list`, it's probably okay as-is.

arcadia_pycolor/tests/test_palette.py
```diff
@@ -0,0 +1,96 @@
...
+    "name, input, hex_codes",
+    [
+        (
+            "my_palette",
+            [HexCode("white", "#FFFFFF"), HexCode("black", "#000000")],
+            [HexCode("white", "#FFFFFF"), HexCode("black", "#000000")],
+        )
+    ],
+)
+def test_palette_from_hexcode_list(name, input, hex_codes):
```
keithchev, line 16: This looks like another case where `input` and `hex_codes` could be condensed into one variable.

arcadia_pycolor/tests/test_palette.py
```diff
@@ -0,0 +1,96 @@
...
+
+@pytest.mark.parametrize(
+    "name, invalid_input",
+    [
+        ("my_palette", [HexCode("white", "#FFFFFF"), "black"]),
+        ("my_palette", [HexCode("white", "#FFFFFF"), 123456]),
+        ("my_palette", [HexCode("white", "#FFFFFF"), 0.123456]),
+    ],
+)
+def test_palette_from_hexcode_list_invalid_input(name, invalid_input):
```
keithchev, line 28: If a list of invalid hex codes is defined as a constant in `test_hexcode`, it could be imported here and also used in this test.

arcadia_pycolor/tests/test_palette.py
```diff
@@ -0,0 +1,96 @@
...
+            [HexCode("white", "#FFFFFF")],
+            [HexCode("black", "#000000")],
+            [
+                HexCode("white", "#FFFFFF"),
+                HexCode("black", "#000000"),
+            ],
+        )
+    ],
+)
+def test_palette_add(name, input, other, result):
```
keithchev, line 95: This is another case where `other` and `result` could be condensed, since `result` is trivially generated from `input + other`.

keithchev (changes requested):
This looks great! I made a bunch of comments but most of them minor. 

General comments (I made inline comments related to these, so this is more just to summarize):
- when there's only one set of parameters and no near-term need for more, it's okay to just defined input and expected values in the bodies of the test function (and have the test function itself have no arguments).
- When possible, eliminate redundant or duplicated information in test parameters, since it can be hard to maintain and it slightly obscures what parameters the test truly depends on.

> How should we test plotting functions, which live in plot.py?

I wouldn't worry about testing these, tbh. It sounds like they're mostly meant for internal power users (e.g. you and Audrey), and they're easy enough to test manually by looking at them. Also, UI tests are just hard - they're brittle and annoying to maintain. I would say we can revisit this in the event that the plotting functions become really important and truly public-facing.

> How do we get GitHub Actions to run the tests? I think I can try to reverse engineer by looking at how it's implemented in ProteinCartography but my brain is too melted to do this today.

This should be easy; it can be done by adding another step to call `pytest .` in the existing `lint.yml` workflow (which we can rename to `ci.yml` or something). I can do this on Monday.

arcadia_pycolor/tests/test_import.py
mezarque: I'd like to include this test just because I've had lots of trouble properly using the import syntax (and importing tends to be a reasonable integration test). This would help catch circular import errors, etc.

mezarque:
Thanks for all the tips! I addressed as many of these comments as I could, aiming to deduplicate things.

keithchev (approved):
LGTM!

PR #14, opened 2024-05-29
Adding matplotlib-specific styling functionality

mezarque:
## PR checklist

## Overview

This PR adds in automatic styling functionality for `matplotlib`, giving users the ability to more easily perform style modifications to plots. The modifications occur primarily through two parts of the API:
- `apc.mpl.setup()`:
    - gets about 80% of the way to styling plots properly
    - meant to be called *just once* at the top of a plotting session
    - loads Suisse family fonts, if they are installed
    - registers Arcadia colors, palettes (as `ListedColormaps`), and gradients (as `LinearSegmentedColormaps`) into `matplotlib` with prefix `apc:`
    - loads a `matplotlib` stylesheet (`arcadia_2024.mplstyle`), updating the defaults in `rcParams`.
- `apc.mpl.autostyle()`
    - because `matplotlib` `rcParams` doesn't allow you to control *everything*, this is needed to get the remaining ~20% towards correct style
    - meant to be called *after each plot*
    - capitalizes axis labels
    - capitalizes and styles the legend to be left-justified, with title in "Suisse Int'l Semibold"
    - adds horizontal line separating legend title from entries
    - using the `mono` argument, alters font of user-chosen ticklabels to be "Suisse Int'l Mono"
    - using the `cat` argument, removes ticks and capitalizes ticklabels for categorical axes (e.g. for violin or bar plots)
    - using the `cbar` argument, alters the font of the tick labels of any colorbars to be "Suisse Int'l Mono"
    - there are functions that perform each of these modifications individually, for power users who want to change things *a la carte*
    
In addition to those major functions, there's some auxiliary functionality to help figures conform to fixed size guidelines, which are provided via existing templates in our Adobe Creative Cloud library.
- `apc.dimensions()`
    - gives a tuple of a fixed dimension, to be passed to `plt.subplots(figsize = <>)`
    - sizes are `"full_w"`, `"full_s"`, `"threeq_w"`, `"threeq_s"`, `"half_s"`
- `apc.SAVE_WEB` and `apc.SAVE_PRINT`
    - provides a collection of arguments to be passed to `plt.savefig(**apc.SAVE_WEB)` to make sure margins, etc. are of the correct size

Drive-by (is this the proper usage of this? I can't find references online):
- split up the `classes.py` module into one module for each class, `hexcode.py`, `gradient.py`, and `palette.py`
    
There's a lot going on here, and I'm sure we'll have a lively discussion about things like naming, what to expose to users, organization, namespaces, etc. Looking forward to lots of feedback!

## Testing
I confirmed that things work as expected by testing the functionality in `usage_example.ipynb`, which goes into some detail about what all is going on and how users might use the various functions.

## Halp
CI is failing right now; it seems that `impresources` is not finding the `arcadia_2024.mplstyle` file and is returning `None`. I am not sure how to resolve this, and would love some help figuring out how to get the CI to work properly; this error does not show up when I run `pytest .` on my local clone.

## TODO
These are things that I'd like to figure out, but haven't yet found an elegant way to implement.
- [ ] Write tests
    - I think there's some base functionality to test and handle, e.g.:
        - What happens if Suisse fonts are not found?
        - What happens if users accidentally call the same function repeatedly?
        - How do we make sure we don't over-register info into `matplotlib` and detect collisions there?
    - Probably, as previously discussed, it will be very hard to test the visualization aspect without just manually looking at plots. It would be nice to have some kind of CI-based approach to generating the plots, though.
- [ ] Provide a declarative utility for adjusting the marker sizes of the legend.
    - When generating a figure using `seaborn`, the marker sizes of legend elements can be really small (equivalent to the size of points in the scatter plot). This is bad b/c it's hard to tell what is a scatter plot point vs. a legend marker. There are [lots of ugly ways](https://stackoverflow.com/questions/48694620/how-to-set-legend-marker-size-and-alpha) to solve this problem, some of which involve just creating a new legend, but it's kind of annoying. Would be nice to have a declarative way to do this, e.g. `apc.mpl.scale_legend_markers(2)` or something.


arcadia_pycolor/mpl.py
```diff
@@ -1,18 +1,336 @@
...
+    SAVEFIG_PAD,
+    THREEQ_S,
+    THREEQ_W,
+)
 
-def palette_to_cmap(palette: "Palette"):
-    return mcolors.ListedColormap([color.hex_code for color in palette.colors], palette.name)
+_DARWIN_FONT_FOLDER = "~/Library/Fonts"
+_FONT_FILTER = "Suisse"
+_MPL_STYLESHEET = impresources.files(mplstyles) / "arcadia_2024.mplstyle"
```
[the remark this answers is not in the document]
mertcelebi: Otherwise, you probably could use something like this to load the file: https://github.com/lgienapp/aquarel/blob/114ba2796d2061dd0ac841c0e982ad0ae2ec0c99/aquarel/utils.py#L38

arcadia_pycolor/mplstyles/arcadia_2024.mplstyle
```diff
@@ -0,0 +1,131 @@
...
+
+### AXES
+axes.facecolor:         none
+axes.edgecolor:         black       # axes edge color
+axes.linewidth:         1           # edge line width
+
+axes.grid:              False       # display grid or not
+axes.grid.axis:         both        # which axis the grid should apply to
+axes.grid.which:        major       # grid lines at {major, minor, both} ticks
+axes.prop_cycle:        cycler('color', ['5088C5', 'F28360', 'F7B846', '97CD78', '7A77AB', 'F898AE', '3B9886', 'C85152', '73B5E3', 'BAB0A8', '8A99AD', 'FFB984', 'C6E7F4', 'F8C5C1', 'F5E4BE', 'B5BEA4', 'DCBFFC', 'B6C8D4', 'DAD3C7', 'DA9085', 'F5CBE4', 'D1EADF', 'BABEE0', 'F1E8DA'])
```
mertcelebi, line 27: what are these colors?

arcadia_pycolor/mplstyles/arcadia_2024.mplstyle
```diff
@@ -0,0 +1,131 @@
...
+font.size:              24
+font.serif:             Suisse Int'l Serif, Times, serif
+font.sans-serif:        Suisse Int'l, Arial, sans-serif
+font.monospace:         Suisee Int'l Mono, Courier, monospace
+font.weight:            regular
+
+### FIGURE
+figure.titlesize:                   26          # size of the figure title (``Figure.suptitle()``)
+figure.titleweight:                 medium      # weight of the figure title
+figure.facecolor:                   FDF8F2      # displays figures with this color
```
mertcelebi, line 14: is this the background color?

arcadia_pycolor/mplstyles/arcadia_2024.mplstyle
mertcelebi: ideally these would come from the colors file

mertcelebi (commented):
I haven't fully reviewed the `mpl` changes, but I made a couple suggestions to start a discussion. I am not sure how much research you've done around available tools, but I would check these two tools as potential inspiration: https://github.com/raybuhr/pyplot-themes and https://github.com/lgienapp/aquarel

arcadia_pycolor/mplstyles/arcadia_2024.mplstyle
mezarque: These are colors from the palette, just input manually. I think moving to a Python dictionary would allow us to import them instead of writing them out like this, which is preferable. I'll get on that!

arcadia_pycolor/mplstyles/arcadia_2024.mplstyle
mezarque: Yep; also should probably be imported from the colors module instead of written out.

arcadia_pycolor/mpl.py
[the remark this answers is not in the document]
mezarque: I think because there were so many parameters that needed adjusting, I used stylesheets, but I think moving to dictionaries would be better, as it keeps everything in Python and allows us to import values from other modules.

mezarque:
> I haven't fully reviewed the `mpl` changes, but I made a couple suggestions to start a discussion. I am not sure how much research you've done around available tools, but I would check these two tools as potential inspiration: https://github.com/raybuhr/pyplot-themes and https://github.com/lgienapp/aquarel

Thanks for pointing me to these! I do like how `aquarel` uses `with` to specify a context for plotting; this could save users from having to have to call `autostyle` at the end of plots. I think both packages do some of the things we need to accomplish, but not all. Configurability/ usability is also something we want to consider, since I think users may not know what changes they want to make in advance of generating a plot. Let's explore these different approaches through some user testing.

arcadia_pycolor/fonts.py
```diff
@@ -0,0 +1,8 @@
+import os
```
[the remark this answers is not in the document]
mezarque: I did not know about this! Let's use this instead.

keithchev (changes requested):
As we discussed offline, I think this is actually most of the way there. Most of my comments are about naming and readability, including stuff I noticed when I tried to use this branch yesterday to style heatmap figures for the peptigate pub.

General comments:
- I think it could improve clarity of the function names in `mpl` all started with verbs. So e.g. instead of `categorical_axes` use `set_axes_style_categorical` or `set_axis_style_to_categorical`.
- It would also help to use consistent nomenclature for functions that set or adjust styles. e.g. `set_axis_style`, `set_legend_style`, etc.

arcadia_pycolor/mpl.py
```diff
@@ -1,18 +1,336 @@
...
+
+def monospace_yticklabels(font: str = _MONOSPACE_FONT, axis=None):
+    "Set the font of the ytick labels to a monospace font."
+    ax = _catch_axis(axis)
+
+    ytick_labels = ax.get_yticklabels()
+    [i.set_fontfamily(font) for i in ytick_labels]
+
+
+def monospace_ticklabels(font: str = _MONOSPACE_FONT, axis=None):
```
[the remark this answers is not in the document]
mezarque: We can change the function name for the general case (accepts any fonts) to better reflect that behavior. But I think it would be good to have a narrower use case, which doesn't accept font arguments, so that users don't have to wonder about the defaults of `set_ticklabels_font`, which by name would not obviously indicate that the font will be changed to a monospaced font.

arcadia_pycolor/mpl.py
```diff
@@ -1,18 +1,336 @@
...
+    line_area = DrawingArea(
+        width=bbox_length, height=0, xdescent=0, ydescent=0
+    )  # width and height in pixels
+    line_area.add_artist(line)
+
+    # Modify the transform of the line to match the DrawingArea's internal coordinate system
+    line.set_transform(line_area.get_transform())
+
+    # Insert the line as a new row just below the title
+    legend_vpacker = legend._legend_handle_box.get_children()[0]
```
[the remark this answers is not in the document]
mezarque: A [VPacker](https://matplotlib.org/stable/api/offsetbox_api.html#matplotlib.offsetbox.VPacker) object sets the vertical spacing around a list of matplotlib Artists. I will double check that the object that's grabbed by this is actually a VPacker, but also am open to renaming this variable.

arcadia_pycolor/gradient.py
```diff
@@ -0,0 +1,77 @@
...
+
+        return "\n".join(
+            [self.swatch()]
+            + [
+                f"{color.swatch(min_name_width=longest_name_length)} {value}"
+                for color, value in zip(self.colors, self.values)
+            ]
+        )
+
+    def to_mpl_linear_cmap(self):
```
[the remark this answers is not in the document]
mezarque: Do we want to override the `to_mpl_cmap` method of the parent class? I suppose it's not likely that someone would want to make a `Gradient` into a `ListedColormap` instead of a `LinearSegmentedColormap`, but we should affirmatively decide one way or the other.

arcadia_pycolor/mpl.py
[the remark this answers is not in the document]
keithchev: Would it make sense to have both a general function `set_ticklabels_font` that requires an axis and a font name (and doesn't provide a default) and then separately a `set_ticklabels_monospace_font` that only accepts an axis?

arcadia_pycolor/mpl.py
```diff
@@ -1,18 +1,336 @@
...
+        if isinstance(grad, Gradient):
+            if (gradname := "apc:" + grad.name) not in colormaps:
+                plt.register_cmap(name=gradname, cmap=grad.to_mpl_linear_cmap())
+
+
+def _load_palettes():
+    for pal in arcadia_pycolor.palettes.__dict__.values():
+        if isinstance(pal, Palette):
+            if (palname := "apc:" + pal.name) not in colormaps:
+                plt.register_cmap(name=palname, cmap=pal.to_mpl_cmap())
```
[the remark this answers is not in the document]
keithchev: This will be even more true if we can rename `Gradient.to_mpl_linear_cmap` to `to_mpl_cmap`

arcadia_pycolor/mpl.py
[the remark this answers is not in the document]
keithchev: I'm so glad I asked 😵‍💫. This stuff feels so opaque/confusing that I think it'd be good to add a comment here to explain why we expect (or know?) that `legend_handle_box.get_children()[0]` is a `VPacker` object and also why that's the object into whose children we need to insert `line_area`.

arcadia_pycolor/gradient.py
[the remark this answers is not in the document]
keithchev: Yeah, I would say we should override `to_mpl_cmap`, because I think we should be shielding the user from the complexity of matplotlib whenever possible (e.g. in this case the difference between `Listed` vs `LinearSegmented`). In the rare case that the user does need a ListedColormap of a gradient, I think it's okay that they have to call the matplotlib function itself.

arcadia_pycolor/mplstyles/arcadia_2024.mplstyle
mezarque: The plots are saved with a transparent background, but this sets the background when the figure is displayed in interactive contexts, e.g. Jupyter notebooks. I think making sure that this color matches with the final style guide will be useful, to that users can see how their chosen colors match to the background.

arcadia_pycolor/mpl.py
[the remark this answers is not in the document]
mezarque: It's a mess lmao. I've added a very verbose description of everything going on; hopefully it saves someone in the future from having to try to figure all of this out.

arcadia_pycolor/mpl.py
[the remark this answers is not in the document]
mezarque: done!

arcadia_pycolor/styles.py
```diff
@@ -0,0 +1,7 @@
+# Units in inches when dpi is 72
+SAVEFIG_PAD = 0.41
+FULL_W = (20.8333, 10.6161)
```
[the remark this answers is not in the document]
mezarque: Added some comments to explain what's going on!

mezarque:
I think I addressed most of the comments above - thanks for all the feedback! My main concern with the renamings is that all of the functions are now quite verbose, which feels clunky / ugly to use. But maybe that's just the sacrifice we have to make for clarity.

Should I write tests for the functionality in this PR, or wait for another PR?

arcadia_pycolor/mpl.py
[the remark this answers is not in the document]
mezarque: I've split this into two functions as recommended, with the second function being `set_ticklabel_monospaced` so that it's slightly less cumbersome to type.

arcadia_pycolor/mpl.py
```diff
@@ -1,18 +1,336 @@
...
-    colors = [(value, color.hex_code) for value, color in zip(gradient.values, gradient.colors)]
-    return mcolors.LinearSegmentedColormap.from_list(
-        gradient.name,
-        colors=colors,
+LEGEND_PARAMS = dict(
+    alignment="left",
+    title_fontproperties={"weight": "semibold", "size": "26"},
+)
+
+SAVE_WEB = dict(dpi=72, bbox_inches="tight", pad_inches=0.41)
```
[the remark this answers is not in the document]
keithchev: Yeah, I think it would be worthwhile to wrap `plt.savefig` so that we can pass in these kwargs and also set the file type, so the user doesn't have to think about this. This can be done and still allow the user to pass in kwargs for `plt.savefig` by doing s/t like this
```python
def save_figure(context="web", **savefig_kwargs):
    kwargs = SAVE_WEB if context == "web" else SAVE_PRINT
    kwargs.update(**savefig_kwargs)
    plt.savefig(**kwargs)
```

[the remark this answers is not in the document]
keithchev: Also would suggest a descriptive name like `SAVEFIG_KWARGS_WEB` instead of `SAVE_WEB`.

arcadia_pycolor/mpl.py
```diff
@@ -1,18 +1,334 @@
...
+
+
+def set_yticklabel_monospaced(axis=None):
+    "Set the font of the ytick labels to a monospace font."
+    ax = _find_axis(axis)
+
+    set_yticklabel_fonts(ax, MONOSPACE_FONT)
+
+
+def set_ticklabel_monospaced(font: str = MONOSPACE_FONT, axis=None):
```
keithchev, line 77: a lingering` font` argument that should be removed

arcadia_pycolor/mpl.py
```diff
@@ -1,18 +1,334 @@
...
+    else:
+        return axis
 
-def gradient_to_linear_cmap(gradient: "Gradient"):
-    colors = [(value, color.hex_code) for value, color in zip(gradient.values, gradient.colors)]
-    return mcolors.LinearSegmentedColormap.from_list(
-        gradient.name,
-        colors=colors,
+
+def set_yticklabel_fonts(axis=None, font: str = DEFAULT_FONT):
```
keithchev, line 39: Nit: there's only one font to set (and ditto for `set_xticklabel_fonts` and `set_ticklabel_fonts`)
```suggestion
def set_yticklabel_font(axis=None, font: str = DEFAULT_FONT):
```

arcadia_pycolor/mpl.py
```diff
@@ -1,18 +1,334 @@
...
+    legend = ax.get_legend()
+
+    if legend is not None:
+        style_legend(legend)
+
+    if categorical_axes == "both":
+        set_xaxis_categorical(axis)
+        capitalize_xticklabels(axis)
+        set_yaxis_categorical(axis)
+        capitalize_yticklabels(axis)
```
keithchev, line 216:
```suggestion
    if categorical_axes == "both":
        set_axes_categorical(axis)
        capitalize_ticklabels(axis)
```

arcadia_pycolor/mpl.py
```diff
@@ -1,18 +1,334 @@
...
+        linestyle="-",
+        transform=line_area.get_transform(),
+    )
+    line_area.add_artist(line)
+
+    # Insert the line as a new row just below the title.
+    # First, get the _legend_handle_box (an HPacker object) which wraps all the legend entries.
+    # Then, get the children of that box, which returns a list of VPacker objects.
+    # Get the first object, which is a VPacker containing the legend entries.
+    # TODO: Check if there are more objects in the case of a multi-column legend.
```
keithchev, line 281: This is v useful context/explanation!

arcadia_pycolor/mpl.py
```diff
@@ -1,18 +1,334 @@
...
+
+def load_colors():
+    "Load Arcadia's colors into the matplotlib list of named colors with the prefix 'apc:'."
+    colors = {"apc:" + color.name: color.hex_code for color in arcadia_pycolor.palettes.all.colors}
+    mpl.cm.colors.get_named_colors_mapping().update(colors)
+
+
+def load_fonts(font_folder: str = None):
+    "Detect and load Suisse-family fonts installed on the system into matplotlib."
+    for fontpath in font_manager.findSystemFonts(fontpaths=font_folder, fontext="ttf"):
```
keithchev, line 304: Would be good to clarify the purpose of `font_folder` here. Does it always need to be provided? Or is it intended only for rare cases where the user hasn't installed the suisse-family fonts system-wide but instead has them in some local directory?

arcadia_pycolor/mpl.py
```diff
@@ -1,18 +1,334 @@
...
+
+def style_axis(axis=None, monospaced_axes=None, categorical_axes=None, colorbar_exists=False):
+    """Apply a set of style changes to a user-specified axis, or (if None), the most recent axis.
+
+    Args:
+        axis (matplotlib.axis.Axis or None): the axis to style
+        monospaced_axes (str): which axes to set to a monospaced font,
+            either 'x', 'y', 'both', or None
+        categorical_axes (str): which axes to set to categorical,
+            either 'x', 'y', 'both', or None
```
keithchev, line 199: Is there additional validation that could/should be done here? It sounds like categorical axes should not be monospaced font, so we might want to check that the x- and y-axes are only either monospaced or categorical. But also feels okay to punt on this for now.

arcadia_pycolor/style_defaults.py
```diff
@@ -0,0 +1,150 @@
...
+THREEQ_S = (7.0904, 6.6946)
+HALF_S = (10.4167, 9.9928)
+
+# Dictionary of common figure sizes.
+FIGURE_SIZES = {
+    "full_w": FULL_W,
+    "full_s": FULL_S,
+    "threeq_w": THREEQ_W,
+    "threeq_s": THREEQ_S,
+    "half_s": HALF_S,
```
keithchev, line 26: Since these keys are the allowed values of `get_figure_dimensions`, they may need to be named something more human-readable if `get_figure_dimensions` will be part of the public API.

keithchev: Also, in either case, I would avoid the single-letter abbreviations "W" and "S", since it's hard to infer what they stand for.

keithchev (changes requested):
Caught a few minor things but very close!

The examples are great! We should (in a later PR) make sure we write a script to programmatically generate them, so we can easily keep them up-to-date.

What happened to the tests? They seem to have disappeared from the CI workflow.

> My main concern with the renamings is that all of the functions are now quite verbose, which feels clunky / ugly to use. But maybe that's just the sacrifice we have to make for clarity.

FWIW, the names don't feel verbose to me, and I feel some verbosity is a necessary sacrifice for clarity. But I'm also perhaps not the most impartial judge 🙈

> Should I write tests for the functionality in this PR, or wait for another PR?

I would say this can wait for a later PR, since this is looking pretty close as-is.

arcadia_pycolor/mpl.py
mezarque: I added some text in the docstring to perhaps make this more clear.

arcadia_pycolor/mpl.py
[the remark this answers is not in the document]
mezarque: Great, thanks for all these suggestions!

arcadia_pycolor/style_defaults.py
mezarque: I feel like fully writing-out `threequarters` is too much, but I expanded the W and S. Ultimately we should make sure that the style guide's terminology for sizing matches with the terminology used here. I'd advocate for something clearer than "full", "three-quarters", and "half" as sizes, because it begs the question "relative to what?" Probably something simple like "small", "medium", "large" and keywords like "tall", "wide", and "square" would be more intuitive.

arcadia_pycolor/mpl.py
mezarque: There may be cases where the categorical axes could be monospaced, e.g. in the case of a sample or strain ID that would be most readable in a monospaced font. So, I think it's fine to allow users to set these on an overlapping basis.

mezarque:
> What happened to the tests? They seem to have disappeared from the CI workflow.

I'm not sure where you're looking, but at least in my view of the PR, the tests are still running (and if you look at the Actions tab, it seems like they're working still?)

I addressed the remaining comments; lmk if you have any further Qs!

arcadia_pycolor/style_defaults.py
keithchev: Agreed. For now, the new names looks clear to me.

keithchev (approved):
Looks good! 🥳 

(Not sure how I thought the tests went messing, but I see them now.)

PR #19, opened 2024-06-03
Remove old CI workflow

keithchev:
This PR removes an old CI workflow and renames the current CI workflow to reflect that it combines linting and testing in one workflow. (Previously, both the old and current workflows were name "lint", which was confusing).


mezarque (approved):
Looks good to me!

PR #21, opened 2024-06-13
Adding a handful of example plotting notebooks

mezarque:
## PR checklist

## Overview
This PR adds in a handful of plotting examples in Jupyter notebooks in the `plot_testing` directory. It also makes adjustments to the `matplotlib` default styles in `style_defaults.py` that bring us closer to the final values (but these will need to be updated again later).

## Testing
All of the notebooks seem to work fine and generate relevant plots correctly.

## Next
After this PR, there will be a separate PR that updates the base gradients, styles, and figure sizing to match the finalized version of the style guide.


arcadia_pycolor/style_defaults.py
```diff
@@ -36,17 +36,22 @@
 MONOSPACE_FONT = "Suisse Int'l Mono"
 
 
+# Specifications for categorical axes.
+CATEGORICAL_AXIS_TICKLENGTH = 0
+CATEGORICAL_AXIS_TICKPADDING = 7
```
mezarque, line 41: I added these variables so that all the style defaults can be in one document, rather than hard-coded in multiple places.

keithchev (approved):
This looks good! I have some comments/thoughts below, but in the interest of expediency, I think it's fine to merge this as-is and address/discuss my comments in later PRs. 

- I feel like `usage_example` notebook has grown a bit too long and may be overwhelming for the reader, so it could be split into multiple notebooks - maybe one for colormaps, one for apc.setup + basic example plots, and one for the CVD utils. 
- The notebooks depend on sklearn, pandas, and seaborn; these should be added to `dev.yml`  (but not to pyroject.toml since they aren't dependencies of the package).
- When the fonts are not installed, there is an avalanche of warnings like this:
  ```
  findfont: Generic family 'sans-serif' not found because none of the following families were found: Suisse Int'l
  ```
  This warning is confusing because I thought there should always be a "sans-serif" fallback font. In any case, we should try to suppress these warnings, if we can, just because they are an unpleasant user experience in a notebook context.

mezarque:
Thanks for this feedback! I created these issues to deal with in future PRs:
- #22 
- #23 
- #24

PR #26, opened 2024-06-18
Fix type hints and add type checking

keithchev:
The pycolor codebase uses type hints but they were incomplete. This PR makes them complete, adds pyright for static type checking, and fixes additional type errors that pyright flagged. I started doing this because I thought it would be easy/quick. It turned out to be less easy and less quick than I imagined, but I got far enough along that I figured it was worth finishing.

The logic of using a static type checker is that if we're going to use type hints, they should be complete and correct, and to know they're correct, we need a type checker. Which means a PR like this, sooner or later. At the same time, I know we didn't talk about this or plan for it, so also happy to punt this for later.

## Drive-bys
- remove the `strict` kwarg from `zip` calls because it was added after python 3.9, which we support.
- add a base class for `Palette` and `Gradient` classes (this was necessary to eliminate type errors).
- add some function overloads where necessary (also to eliminate type errors).
- some minor variable renaming for clarity.
- remove the now-unused `functions.py` module.

## Notes
- There are no type hints for the tests; I figured this was less important and could wait for later.
- Matplotlib types are incomplete. This leads to a lot of `# type: ignore` comments when internal matplotlib objects are used. afaict, there's no way around this.
- Types for matplotlib were added in 3.8. We support 3.7, so to run type checking locally or in CI, we'll need to use at least 3.8.


arcadia_pycolor/cvd.py
```diff
@@ -47,22 +58,21 @@ def simulate_color(
...
 
-    returned_colors = []
-    for color in processed_colors:
+    returned_colors: list[HexCode] = []
+    for color in colors:
         rgb_color = color.to_rgb()
         cvd_color_name = f"{color.name}_{cvd_type}"
-        cvd_rgb_color = np.clip(cspace_convert(rgb_color, cvd_space, "sRGB1") / 255, 0, 1)
-        cvd_hexcode = HexCode(name=cvd_color_name, hex_code=mpl.colors.to_hex(cvd_rgb_color))
+        cvd_color = cast(NDArray[np.int64], cspace_convert(rgb_color, cvd_space, "sRGB1"))
```
mezarque, line 67: Just curious, do we need to use `cast` for calls to *any* package that doesn't use typing in its implementation? Is that also why there are the `# type: ignore` tags after every `mpl` call?

arcadia_pycolor/cvd.py
```diff
@@ -47,22 +58,21 @@ def simulate_color(
...
+        cvd_color = np.clip(cvd_color / 255, 0, 1)
+        hex_code = mpl.colors.to_hex(cvd_color)  # type: ignore
+        cvd_hexcode = HexCode(name=cvd_color_name, hex_code=hex_code)
         returned_colors.append(cvd_hexcode)
 
     if len(returned_colors) == 1:
         return returned_colors[0]
-    else:
-        return returned_colors
+    return returned_colors
```
mezarque, line 75: Is there a convention for if-else return statements to be structured like this?

arcadia_pycolor/palette.py
```diff
@@ -1,16 +1,21 @@
...
 
 from arcadia_pycolor.display import colorize
 from arcadia_pycolor.hexcode import HexCode
 
+T = TypeVar("T", bound="ColorSequence")
+
 
-class Palette:
-    def __init__(self, name: str, colors: list[HexCode]):
+class ColorSequence(abc.ABC, Generic[T]):
```
mezarque, line 12: I'm curious what defining an abstract base class gives us that is better than what having `Gradient` inherit from `Palette`. Is this just the convention for typing related classes?

arcadia_pycolor/hexcode.py
```diff
@@ -34,23 +35,41 @@ def __new__(cls, name: str, hex_code: str):
         obj.hex_code = hex_code
         return obj
 
-    def to_rgb(self):
+    @property
+    def name(self) -> str:
+        return self._name
+
+    @name.setter
```
mezarque, line 42: Is the addition of `setter` and `property` decorators because we want those attributes of the object to be non-mutable?

arcadia_pycolor/mpl.py
```diff
@@ -1,10 +1,13 @@
...
+
 import matplotlib as mpl
 import matplotlib.font_manager as font_manager
 import matplotlib.pyplot as plt
-from matplotlib import colormaps
+from matplotlib import colormaps  # type: ignore
 from matplotlib.legend import Legend
 from matplotlib.lines import Line2D
 from matplotlib.offsetbox import DrawingArea
+from matplotlib.pyplot import Axes  # type: ignore
```
mezarque, line 10: We use `# type: ignore` here, but we also use `Axes` as part of type hints later. So, does this mean we just use `Axes` as a type that we want to check, but Pyright doesn't try to look under the hood of what `Axes` actually contains?

mezarque (approved):
Just a couple of questions for my understanding, but otherwise this looks good to me (as far as I can understand it).

arcadia_pycolor/cvd.py
keithchev: Yes and yes. People solve this problem by writing what are called type stubs (which specify types for an untyped package and can be written and distributed separately from the package itself). From what I can gather, Matplotlib is in an unhappy middle ground; it now has some types, so people aren't writing type stubs for it, but they're incomplete, so we're forced to use a lot of `type: ignore` flags. And afaict no one's written type stubs for colorspacious, so we're also out of luck there.

arcadia_pycolor/cvd.py
keithchev: afaik there's no particular convention about this, other than that generally it's best to minimize complexity, and reducing the number of indented blocks is one way to do this. I think it's also considered a good practice to make sure functions have at least one unconditional return statement to avoid the possibility that subsequent changes made to the if-else block introduce a bug that results in the function sometimes returning nothing. (An even more strict best practice is that functions should have only one unconditional return statement, with intermediate steps handled by assignment, which is a bit easier to read and lint, but that feels like overkill here).

arcadia_pycolor/palette.py
keithchev: `Gradient` can't inherit from `Palette` because Pyright in "strict" mode* requires that overridden methods must have compatible signatures. In this case, that means that if `Palette.__add__` accepts an argument of type `Palette` then `Gradient.__add__` can't accept an argument of type `Gradient` because that type is more specific than (i.e. a subtype of) the `Palette` type. The logic of this rule is to ensure that it is possible to use a subclass of a parent class in place of its parent class anywhere the parent class is used (TIL this is called the Liskov substitution principle).

In this case, the easiest way around this seemed to be to define a shared base class for Palette and Gradient. I felt this also made conceptual sense, since arguably a gradient is not conceptually an extension of a palette so much as it is a distinct type of color sequence.

*ironically I ended up giving up on "strict" mode, so technically we could leave this as it was, but by the time I realized "strict" mode was too much trouble, I had already made this change.

arcadia_pycolor/hexcode.py
keithchev: This is to address another typing error. If I'm remembering correctly, in a "normal" class, pyright determines the public instance attributes from assignments like `self.name = name`, but because there's no `__init__` method for subclasses of `str`, we have to declare the instance attributes as properties so that pyright "knows" what the public instance attributes are.

arcadia_pycolor/mpl.py
keithchev: I'm not totally sure but I *think* the type error that's ignored here is related to the fact that `Axes` isn't explicitly declared as an export of `matplotlib.pyplot`, which breaks a pyright rule. But obviously `Axes` is defined and *is* importable (because python has no actual implementation of private vs public anything) and so pyright can follow the import and use `Axes` later on to type check the `Axes`-typed objects.

arcadia_pycolor/cvd.py
mezarque: 😢

arcadia_pycolor/palette.py
mezarque: Thanks for explaining all of this! All this rationale makes sense to me.

arcadia_pycolor/hexcode.py
mezarque: I see, thanks for explaining!

arcadia_pycolor/mpl.py
mezarque: Thanks for explaining!

mertcelebi (approved):

Issue #23, opened 2024-06-14
Update `dev.yml` to include example notebook dependencies


mezarque:
> The notebooks depend on sklearn, pandas, and seaborn; these should be added to dev.yml (but not to pyroject.toml since they aren't dependencies of the package).

keithchev:
Addressed in #28.

PR #28, opened 2024-06-26
Add plotting tests

keithchev:
This PR adds integration-like tests that generate realistic plots of various common kinds. They are derived from the notebooks in `plot_testing/` and are intended to replace them (so this PR also deletes the notebooks).

This PR also adds a pytest CLI option to save the PDF files generated by these tests in a user-specified directory. This allows for manual inspection of the plots generated by the tests. See the README for details.

## Drive-bys
- fix a bug introduced in #26 in the colormap registration function.
- disallow matplotlib 3.8.0 because it has a bug that breaks `apc.mpl.setup`.
- disallow matplotlib 3.9 because it introduces breaking API changes.
- add missing dev deps (seaborn, sklearn, pandas) (addresses #23).
- update the poetry lockfile.


arcadia_pycolor/mpl.py
```diff
@@ -347,17 +346,22 @@ def load_colormaps() -> None:
...
+        object
+        for object in (
+            list(arcadia_pycolor.palettes.__dict__.values())
+            + list(arcadia_pycolor.gradients.__dict__.values())
+        )
+        if isinstance(object, ColorSequence)
+    ]
+
+    for colormap in colormaps:
+        if (colormap_name := f"apc:{colormap.name}") not in colormaps:
```
mezarque, line 359: Previously, this conditional statement was checking whether the `apc` colormap was located in `matplotlib`'s `colormaps` dictionary. Now, the imported `matplotlib` `colormaps` variable has been reassigned to a `colormaps` list generated above, which I believe means we're no longer checking whether the colormaps are already registered or not.

arcadia_pycolor/tests/plotting/test_barplots.py
```diff
@@ -0,0 +1,210 @@
...
+    100000,
+    110000,
+    20000,
+    60000,
+    120000,
+    50000,
+]
+
+
+def _add_commas_to_axis_tick_labels(axis):
```
mezarque, line 61: This is an esoteric but useful enough utility that I wonder if we should include it in the package...

arcadia_pycolor/tests/plotting/test_barplots.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    Add commas to the numbers used for axis tick labels.
+    """
+    axis.set_major_formatter(matplotlib.ticker.FuncFormatter(lambda x, _: format(int(x), ",")))  # type: ignore
+
+
+def _add_line_returns_to_sample_ids(sample_ids):
```
mezarque, line 68: Nit: maybe `_word_wrap_sample_ids`?

pyproject.toml
```diff
@@ -32,14 +32,18 @@ url = "https://test.pypi.org/legacy/"
 [tool.poetry.dependencies]
 colorspacious = ">=1.1.2"
 ipykernel = ">=6.25.2"
-matplotlib = ">=3.7.0"
+# There is a bug in `LinearSegmentedColormap` in matplotlib 3.8.0.
+matplotlib = ">=3.7,<3.9,!=3.8.0"
```
mezarque, line 36: We should probably try to avoid making this package this tightly constrained by version numbers in the future.

mezarque (changes requested):
Thanks for implementing these tests! Left some thoughts, including about one change that (I think) introduces a breaking bug.

One higher-level question: should we remove the Jupyter notebooks? The goal of those notebooks is to serve as example code for folks to be able to understand how the library can be used in different contexts. Replacing those entirely with the tests prevents users from being able to have more interpretable examples to look at. We don't have to keep those exact notebooks or put them in the same place, but I do think it would be good for us to provide some copypaste-able code for this.

pyproject.toml
keithchev: Totally. I think 3.8.0 we have to disallow indefinitely, but the `<3.9` was necessary just for now to get the CI to pass (otherwise 3.9 is installed, since it's the newest).

arcadia_pycolor/tests/plotting/test_barplots.py
keithchev: Yeah, can't hurt. Moved this to `apc.mpl`.

arcadia_pycolor/mpl.py
keithchev: Good catch! Fixed and also added an implicit test for this by calling `apc.mpl.setup` multiple times.

keithchev:
> One higher-level question: should we remove the Jupyter notebooks? The goal of those notebooks is to serve as example code for folks to be able to understand how the library can be used in different contexts. Replacing those entirely with the tests prevents users from being able to have more interpretable examples to look at. We don't have to keep those exact notebooks or put them in the same place, but I do think it would be good for us to provide some copypaste-able code for this.

Yeah, makes sense. For now I've added the notebooks back into the `plot_testing/` directory but we can move these to a more appropriately named directory later on.

mezarque (approved):
We did it, Joe

Issue #41, opened 2024-07-12
Gradient color interpolation


ekiefl:
I'm using nglview and would like to register custom color gradients for coloring protein surfaces. Unfortunately, you can't pass a matplotlib colormap or a custom gradient definition like in plotly (`[(c1, v1), ...]`). This means I need more granular control, and must interpolate colors across the gradient based on a value range.

I think this could be part of the `Gradient` class. Something like:

```python
def interpolate_color(
    self, values: Iterable[float], min_value: float, max_value: float
) -> list[float]:
    if not all(min_value <= value <= max_value for value in values):
        raise ValueError("Value must be within the range [min_value, max_value].")

    cmap = self.to_mpl_cmap()
    return [mcolors.to_hex(cmap((value - min_value) / (max_value - min_value))) for value in values]
```

Semantically, I guess it should really return a list of `HexCode` objects, but practically I think all use cases would want the hex strings directly.

I guess it would also be better to snap any values outside the range to the closest extremum.

mezarque:
Thanks for this suggestion! I'm curious whether this needs to be within the package, or whether the mapping should instead exist be managed outside of this package, in the code that you're using to interface with `nglviewer`.

If you're looking for a list of HexCode objects that form a set number of bins within your defined gradient, you can resample the gradient using `Gradient.resample_as_palette(steps)`. This returns a Palette of HexCodes with a number of colors equal to `steps`, which you can then map onto your ranges using `zip` or an equivalent function.

ekiefl:
> I'm curious whether this needs to be within the package, or whether the mapping should instead exist be managed outside of this package, in the code that you're using to interface with nglviewer

As the person suggesting this feature, I'll be the first to say it doesn't _need_ to exist in this package. But I believe sampling a color from a gradient based on a value range is the meat and potatoes of what a gradient is used for. That's what happens when we pass an argument for `cmap=` in matplotlib, and the operation is so ubiquitous that it's even implemented as the very `__call__` method of matplotlib's `Colormap`.

While I can implement this as a utility in my own code, I don't see my use case as bespoke. I see it as belonging to a broad class of use cases when users can't hand over the responsibility of mapping values to colors to packages like matplotlib and seaborn, which directly ingest the `Colormap` generated by `to_mpl_cmap`. Including it in my code feels like it is fulfilling a responsibility that ought to be fulfilled by our color package.

> If you're looking for a list of HexCode objects that form a set number of bins within your defined gradient

[...]

PR #43, opened 2024-07-16
Add map_values to Gradient

ekiefl:
# Add map_values method to Gradient

Proposed solution to https://github.com/Arcadia-Science/arcadia-pycolor/issues/41

Something that gave me pause: on one hand I want to "buy in" to the existing data structures and make the return type `list[HexCode]`, but on the other hand I feel it's more practical to return the hex strings since I suspect any callers of this function are typically using it as a final step in their process with the library. I'm curious what you think @mezarque.


arcadia_pycolor/gradient.py
```diff
@@ -94,6 +95,53 @@ def resample_as_palette(self, steps: int = 5) -> Palette:
             colors=colors,
         )
 
+    def map_values(
```
keithchev, line 98: I think for consistency it would be good to have this method return `list[HexCode]`, since `HexCode` is intended to be how this package always represents strings-that-are-hex-codes. Also, because `HexCode` is a subclass of `str`, I don't *think* this would have any effect on users/calling functions.

arcadia_pycolor/gradient.py
```diff
@@ -94,6 +95,53 @@ def resample_as_palette(self, steps: int = 5) -> Palette:
...
+            )
+
+        cmap = self.to_mpl_cmap()
+
+        if min_value == max_value:
+            # Value range is 0. Return the midrange color for each value.
+            return [mcolors.to_hex(cmap(0.5))] * len(values)
+
+        normalized_values = [(value - min_value) / (max_value - min_value) for value in values]
+        clamped_values = [max(0.0, min(1.0, value)) for value in normalized_values]
```
mezarque, line 141: I wonder if we should warn the user when values outside of the min/max range are provided, rather than clamping those values silently here.

keithchev (commented):
Just driving by and thought I'd drop some comments (I'll leave this for @mezarque to approve since it's related to the public-facing API).

arcadia_pycolor/gradient.py
```diff
@@ -94,6 +95,53 @@ def resample_as_palette(self, steps: int = 5) -> Palette:
...
+        values: NumericSequence,
+        min_value: Union[float, None] = None,
+        max_value: Union[float, None] = None,
+    ) -> list[str]:
+        """Map a sequence of values to their corresponding colors from a gradient
+
+        Args:
+            min_value:
+                Determines which value corresponds to the first color in the spectrum.
+                Values less than this are given this color. If not provided, min(values) is
```
mezarque, line 109: Phrasing feels a bit awkward. Maybe "Values below the minimum are assigned to the minimum color"?

mezarque (changes requested):
Thanks for implementing this! Took a look and left some suggestions.

arcadia_pycolor/gradient.py
```diff
@@ -94,6 +95,53 @@ def resample_as_palette(self, steps: int = 5) -> Palette:
...
+            max_value = max(values)
+
+        if min_value > max_value:
+            raise ValueError(
+                f"max_value ({max_value}) must be greater than min_value ({min_value})."
+            )
+
+        cmap = self.to_mpl_cmap()
+
+        if min_value == max_value:
```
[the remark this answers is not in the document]
mezarque: +1

arcadia_pycolor/gradient.py
```diff
@@ -94,6 +95,53 @@ def resample_as_palette(self, steps: int = 5) -> Palette:
             colors=colors,
         )
 
+    def map_values(
+        self,
+        values: NumericSequence,
+        min_value: Union[float, None] = None,
+        max_value: Union[float, None] = None,
+    ) -> list[str]:
```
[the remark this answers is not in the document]
ekiefl: I agree. To be honest, my reservation to returning `HexCode` stemmed from me not understanding what the string value was. For some reason I thought the string value of `apc.aegean` was `aegean`, not its hex code. I'm happy I was wrong because this makes much more sense both inside and outside of apc land.

Added in 566b705389626922ce6260cc2451564e1705dee8, however I had to make a decision about the HexCode name that I'd like your feedback on.

image

Personally, I like that the values are accessible, though it is abusing the name attribute to be sure.

arcadia_pycolor/gradient.py
ekiefl: 566b705389626922ce6260cc2451564e1705dee8

See https://github.com/Arcadia-Science/arcadia-pycolor/pull/43#discussion_r1680145395

arcadia_pycolor/gradient.py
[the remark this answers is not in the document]
ekiefl: 👍 8328e6781d6cc92eef33ff465c33887c4cce4d08

arcadia_pycolor/tests/test_gradients.py
```diff
@@ -116,3 +117,52 @@ def test_gradient_swatch_steps():
...
+    "values, expected_colors",
+    [
+        ([0, 1], ["#000000", "#ffffff"]),
+        ([0, 0.5, 1], ["#000000", "#808080", "#ffffff"]),
+        ([1, 2, 3, 4, 5], ["#000000", "#404040", "#808080", "#c0c0c0", "#ffffff"]),
+        ([-1, 0, 1], ["#000000", "#808080", "#ffffff"]),
+        ([], []),
+    ],
+)
+def test_map_values_basic_cases(
```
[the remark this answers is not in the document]
ekiefl: c3edb8285493da686dd1d1bde031223be4e09838

arcadia_pycolor/gradient.py
ekiefl: I can only speak to my personal usage with colormaps, but if I'm touching the min and max values, it's usually because I'm interested in the middle of the color range, but far-reaching extrema values are decreasing contrast.

Like if your data is spread like this (each `x` a datapoint, each `|` the min/max value):

```bash
x                           x x    x x  x x x x x x x x                              x
|                                                                                    |
red                                  yellow                                      green
```

The bulk of the data is washed out. In comparison, by setting min/max inside the value range:

```bash
x                           x x    x x  x x x x x x x x                              x
                          |                             |
red                       red          yellow       green                        green
```

You get more contrast in the region you care about.

All this is to say that I personally would not want a warning that I'm doing this, because it would be very intentional.

arcadia_pycolor/gradient.py
ekiefl: 9f0c6110aec7960bad39a4f65b327542de394db5

arcadia_pycolor/gradient.py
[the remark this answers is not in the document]
mezarque: I think this is fine; we make a similar kind of assignment when using `apc.gradients.resample_as_palette()`. I do feel like rounding the values to a fixed number of digits would be more legible, but I suppose that's not ideal if you want to be able to exactly assign a color to a value.

arcadia_pycolor/gradient.py
mezarque: I understand this use case a bit better, thanks! I do wonder if it would be good to use the `matplotlib` [`set_extremes`](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.Colormap.html#matplotlib.colors.Colormap.set_extremes) functionality when windowing the data in this way, so that colors over and under the gradient are distinctly marked as out of range. Probably not relevant to your structure viewer use case, but in a heatmap or other context, it would be important to indicate so as to avoid misleading readers.

mezarque (approved):
Looks good to me! Thanks for addressing those comments.

Before merging, I do wonder if we should get a sense of how often we want to cut new releases on GitHub and distributions to PyPI. @keithchev any thoughts? I imagine it would make Evan's life a little easier if this is included in the PyPI-distributed version, but we should be disciplined about what amount of change necessitates a minor or patch version bump.

keithchev:
> Before merging, I do wonder if we should get a sense of how often we want to cut new releases on GitHub and distributions to PyPI. @keithchev any thoughts? I imagine it would make Evan's life a little easier if this is included in the PyPI-distributed version, but we should be disciplined about what amount of change necessitates a minor or patch version bump.

Yeah, if we end up with like version `0.5.200` we're probably doing something wrong. And of course we want to be disciplined about releasing breaking changes. But I think we can err on the side of leniency for minor versions for now, at least for new features like this one that may unblock or facilitate a user's work.

mezarque:
Sounds good! Feel free to merge @ekiefl. @keithchev would you mind distributing to PyPI?

arcadia_pycolor/gradient.py
ekiefl: Yeah good idea. By name, I thought that `set_extremes` would allow us to avoid normalizing the data altogether, by setting the extrema for the value range. But looking into the implementation, I now see it allows you to control the colors of those outside 0 to 1.

This could be useful if we wanted more fancy features and could be accomplished by propagating those args to the `map_values` call signature.

Either way, it's noteworthy that the default behavior for `over` and `under` is to clamp the color to the lowest/highest color. Here's a black to white cmap:

```
(Pdb) mcolors.to_hex(cmap(-1.0))
'#000000'
(Pdb) mcolors.to_hex(cmap(2.0))
'#ffffff'
```

That's what we we were manually doing. As a step towards implementing `over`/`under`/`bad`, the clamping op has been removed (0c1c8af9cc3530276271874484599f8c28c03480).

ekiefl:
Merging now. Made a small change described here: https://github.com/Arcadia-Science/arcadia-pycolor/pull/43#discussion_r1681397776

PR #44, opened 2024-07-17
Update dependencies

keithchev:
This PR updates `certifi` to address [this security alert](https://github.com/Arcadia-Science/arcadia-pycolor/security/dependabot/1). It also includes the changes to the lockfile made by `poetry update`. Both of these changes only affect the lockfile.


mezarque (approved):
lgtm!

Issue #40, opened 2024-07-07
"HexCode" in colors_all


ekiefl:
```python
import arcadia_pycolor as apc
assert "HexCode" in apc.colors_all
```

A quick fix, given the current import structure, would be (untested):

```python
colors_all = [name for name in dir(colors) if isinstance(getattr(colors, name), HexCode)]
```

mezarque:
Nice, this seems to work! Thanks for noticing that!

mezarque:
Resolved by #46

PR #46, opened 2024-07-17
Import fixes

keithchev:
This PR addresses #40 and also removes the star imports in the top-level `__init__` in favor of explicit imports (and adds them to `__all__`).


mezarque (approved):
Lgtm!

PR #48, opened 2024-07-17
Add dynamic versioning

keithchev:
This PR introduces dynamic versioning that infers the package version from the most recent git tag. This simplifies the release workflow by eliminating the need for a separate PR to increment a hard-coded version number in `pyproject.toml` or elsewhere. It also takes care of populating the `__version__` attribute in the top-level `__init__.py`. 

With this change, we can create and publish a new version simply by creating a new git tag containing the new version number, then calling `make publish` as before.

Another nice feature of this approach is that the generated version string will include flags if there are commits since the most recent git tag or if the repo contains uncommitted changes (neither of which should ever be the case when a new release is published, but there is no simple way to enforce this, so appending this information in the version string is a nice sanity check). 

h/t to @ekiefl for pointing out that the package was missing a `__version__` attribute, which motivated this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,11 +63,15 @@ Some of the documentation is in the form of Jupyter notebooks. The inline graphi
 
 Run the makefile command `execute-all-notebooks` to execute all the notebooks. This 1) ensures that the notebooks execute without errors and 2) updates their outputs in-place. Then, commit any modified notebooks to the repo.
 
-## Releasing the package on PyPI
+## Publishing a new version of the package on PyPI
 
-Releasing a new version of the package on PyPI requires API tokens for the test and real PyPI servers. You can find these tokens in your PyPI account settings. Copy `.env.copy` to `.env` and add your tokens to this file.
+Publishing the package on PyPI requires that you have API tokens for the test and production PyPI servers. You can find these tokens in your PyPI account settings. Copy `.env.copy` to `.env` and add your tokens to this file.
 
-We use git tags to define versions. When you're ready to release a new version of the package, first create a new git tag. __Make sure your local git repository is up-to-date before creating the tag!__
+To release a new version of the package on PyPI, its version number must first be incremented.
+
+We use git tags to define versions. When you're ready to release a new version of the package, first create a new git tag. The name of the tag should correspond to the version number, e.g. `"v0.1.0"`. Annotate the tag with a message that describes the release, e.g. "Release version 0.1.0". 
+
+__Before creating the tag, make sure that your local git repository is on `main`, is up-to-date, and does not contain uncommitted changes!__
 
 ```bash
 git tag -a v0.1.0 -m "Release version 0.1.0"
@@ -76,7 +80,22 @@ git push origin v0.1.0
 
 We use semantic versioning in which the versions have the form `MAJOR.MINOR.PATCH`. See [here](https://semver.org/) for more information.
 
-Then, build the package and upload the build artifacts to the PyPI test server:
+Next, build the package:
+```bash
+make build
+```
+
+You should see an output that looks like this:
+```
+Building arcadia-pycolor (0.1.0)
+  - Building sdist
+  - Built arcadia_pycolor-0.1.0.tar.gz
+  - Building wheel
+```
+
+Check that the version of the package matches the tag you created; Poetry automatically infers this version from the git tag that you created.
+
+Next, test that you can publish the package to the PyPI test server:
 
 ```bash
 make test-publish
@@ -92,13 +111,13 @@ Check that you can install the package from the test server:
 pip install --index-url https://test.pypi.org/simple/ arcadia-pycolor
 ```
 
-If everything looks good, build and publish the package to the real PyPI server:
+If everything looks good, build and publish the package to the prod PyPI server:
 
 ```bash
 make publish
 ```
 
-Finally, check that you can install the package from the real PyPI server:
+Finally, check that you can install the package from the prod PyPI server:
 
 ```bash
 pip install arcadia-pycolor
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -63,18 +63,39 @@ Some of the documentation is in the form of Jupyter notebooks. The inline graphi
...
+```
+Building arcadia-pycolor (0.1.0)
+  - Building sdist
+  - Built arcadia_pycolor-0.1.0.tar.gz
+  - Building wheel
+```
+
+Check that the version of the package matches the tag you created; Poetry automatically infers this version from the git tag that you created.
+
+Next, test that you can publish the package to the PyPI test server:
```
ekiefl, line 98: I have a question about the Makefile. If you run `build` and subsequently `test-publish`, is the `build` command ran again, given the current rule?

```
.PHONY: test-publish
test-publish: build
	poetry publish \
		--repository pypi_test \
		--username __token__ \
		--password ${POETRY_PYPI_TOKEN_PYPI_TEST}
```

arcadia_pycolor/__init__.py
```diff
@@ -5,6 +5,9 @@
 from .hexcode import HexCode
 from .palette import Palette
 
+# This is a placeholder that will be replaced by the version number at build time.
+__version__ = "0.0.0"
+
```
ekiefl, line 10: Am I correct that anyone running on a developmental branch will have `__version__ == "0.0.0"`? If so, maybe our placeholder could be more informative?

pyproject.toml
```diff
@@ -1,10 +1,22 @@
...
+style = "pep440"
+# Append a flag to the version number if there are uncommitted changes.
+dirty = true
 
 [tool.poetry]
 name = "arcadia-pycolor"
-version = "0.5.1"
+# This version is a placeholder. It will be updated at build time by poetry-dynamic-versioning
+# from the latest git tag.
+version = "0.0.0"
```
ekiefl, line 19: Really cool

ekiefl (commented):
I've never seen this, but I really like it.

Have you tested that it works? I tried installing from test PyPI but the highest version is 0.5.1, and doesn't have a __version__ attribute.

It would be nice to confirm that this works:

```
make build

conda deactivate
conda deactivate
conda deactivate
conda deactivate
conda deactivate
conda env remove -y --name delete
conda create -y -n delete python=3.9
conda activate delete

# Untested
pip install dist/*tar.gz

# Move out of directory to not accidently import locally
cd ~
python -c "import arcadia_pycolor; print(arcadia_pycolor.__version__)"
```

mezarque (approved):
I'm really not familiar with most of poetry's functionality, but otherwise this seems like a useful change!

README.md
keithchev: Oops yeah I don't think it is, but it should be. It looks like `test-publish` should depend on `clean`, not `build`.

keithchev: Double oops. I think it's correct as it is, because `build` depends on `clean`. So calling `test-publish` will always re-build the package.

keithchev:
> Have you tested that it works? I tried installing from test PyPI but the highest version is 0.5.1, and doesn't have a version attribute.

Yeah, I tested it locally by checking that `__version__` gets updated in the `.gz` build artifact. It's not possible to test-publish without a new version, because pypi doesn't allow the "dirty" version numbers that are generated when the package is built from a tag-less commit. For example, building on this PR results in a version "0.5.1.post3.dev0+0a49a94". I like this feature of the dynamic versioning and it makes sense that pypi disallows these versions, since it prevents accidentally publishing some random commit, but it does make testing the publishing command harder...

arcadia_pycolor/__init__.py
keithchev: Yeah, that's correct. This is the placeholder recommended by [the poetry-dynamic-versioning plugin](https://github.com/mtkennerly/poetry-dynamic-versioning) so that's why I chose it. I'll see if it can be changed. What do you think would be more informative?

README.md
ekiefl: I think it's correct, too. It just gave me pause because this cumulative pipeline (build calls clean, publish calls build calls clean) means the `make build` command in the instructions preceding isn't doing any work--it's just a very important sanity check whose output is about to be wiped by the publish command. It's very much not a big deal but I would prefer any of these options over what currently happens:

1. A note is made in the instructions that `[test-]publish` wipes dist/ and rebuilds the package
2. `[test-]publish` does not re-build the package
3. `[test-]publish` is renamed to something like `build-and-[test-]publish` to be more descriptive

arcadia_pycolor/__init__.py
ekiefl: That's reasonable. Since we've committed to the plugin, let's just go all in and adopt their convention.

FYI this sparked me to look at how some big packages populate their `__version__` attribute and well... it gets messy 😱 (https://github.com/pandas-dev/pandas/blob/main/pandas/_version.py)

ekiefl (approved):
Left a comment you should feel free to take or leave.

Thanks for the intro to dynamic versioning.

ekiefl:
I feel like this has been on our minds but since it hasn't yet been expressed yet, allow me to be first to say: this is chalking up to be a pretty robust publishing pipeline poster child that should be adaptable to other packages 🎊

README.md
keithchev: Totally. I guess the implicit idea/assumption is that "building" is a fast operation that generates ephemeral artifacts. But this isn't clear from the nomenclature. I'll rename `test-publish` to `build-and-test-publish` and similarly for `publish`.

arcadia_pycolor/__init__.py
keithchev: 😵‍💫

keithchev:
> I feel like this has been on our minds but since it hasn't yet been expressed yet, allow me to be first to say: this is chalking up to be a pretty robust publishing pipeline poster child that should be adaptable to other packages 🎊

Thanks! It's slowly getting there, though it still feels v minimal compared to "real" packages (e.g. [this](https://github.com/pandas-dev/pandas/blob/main/.github/workflows/wheels.yml) which I *think* is pandas' actual build workflow, though it's hard to tell).

Issue #49, opened 2024-07-30
Add ability to save figure as multiple filetypes simultaneously


reillyowencooper:
I often want to save a .pdf and .png file for each figure I make, which results in a lot of duplicate lines of code where the only thing I've changed is the extension. It would be great to have an option in `save_figure` that allows you to save both filetypes (or more, if people use other extensions) at the same time. Something like:

```
def save_figure(fname: str, fname_types: list = ["pdf, png"], context: str = "web", **savefig_kwargs: dict[Any, Any]) -> None:
    "Save the current figure with the default settings for web."
    kwargs = SAVEFIG_KWARGS_WEB if context == "web" else SAVEFIG_KWARGS_PRINT
    kwargs.update(**savefig_kwargs)  # type: ignore
    for fname_type in fname_types:
        plt.savefig(fname = fname + fname_type, **kwargs) #type: ignore
```

mezarque:
Addressed by #51

PR #51, opened 2024-08-01
Add ability to save multiple different file types

mezarque:
## PR checklist

## Overview

This PR adds:
- Ability to save a single plot as multiple different file types when using `apc.mpl.save_figure`, as suggested by #49. Thanks @reillyowencooper for the feature suggestion!
- Conditional logic for handling when a user provides vs. doesn't provide valid file suffixes.
- Tests that conditional logic is working.
- Small update to `docs/style_usage.ipynb` to demonstrate use case.


arcadia_pycolor/mpl.py
```diff
@@ -62,11 +64,45 @@ def _arcadia_fonts_found() -> bool:
...
+    fname_types: Union[list[str], None] = None,
+    context: str = "web",
+    **savefig_kwargs: dict[Any, Any],
+) -> None:
+    """
+    Save the current figure, accounting for Arcadia's defaults.
+
+    Args:
+        fname (str): the filename to save the figure to
+        fname_types (list, optional): the filetypes(s) to save the figure to.
```
keithchev, line 78: Nit: can this be called `filetypes`? The name `fname_types` is a bit weird, if you think about it.

arcadia_pycolor/mpl.py
```diff
@@ -62,11 +64,45 @@ def _arcadia_fonts_found() -> bool:
...
+            If the original suffix is not in fname_types, it is appended to the list.
+        context (str): the context to save the figure in, either 'web' or 'print'
+        **savefig_kwargs: additional keyword arguments to pass to plt.savefig
+    """
     kwargs = SAVEFIG_KWARGS_WEB if context == "web" else SAVEFIG_KWARGS_PRINT
     kwargs.update(**savefig_kwargs)  # type: ignore
-    plt.savefig(fname=fname, **kwargs)  # type: ignore
+
+    # Gets a list of valid suffixes for saving figures from matplotlib.
+    valid_suffixes = list(FigureCanvasBase.get_supported_filetypes().keys())
```
keithchev, line 88: Nit: use `filetype` instead of `suffixes`
```suggestion
    valid_filetypes = list(FigureCanvasBase.get_supported_filetypes().keys())
```

arcadia_pycolor/mpl.py
```diff
@@ -62,11 +64,45 @@ def _arcadia_fonts_found() -> bool:
...
+        suffix = Path(fname).suffix[1:]  # Slice to remove the period.
+        filepath_no_suffix = Path(fname).with_suffix("")
+
+        # Handle the case where the original suffix is not in fname_types.
+        if (suffix != "") and (suffix not in fname_types):
+            fname_types.append(suffix)
+
+        for suffix in fname_types:
+            if suffix not in valid_suffixes:
+                print(f"Invalid file suffix '{suffix}'. Skipping.")
```
keithchev, line 100:
```suggestion
                print(f"Invalid filetype '{suffix}'. Skipping.")
```

arcadia_pycolor/mpl.py
```diff
@@ -62,11 +64,45 @@ def _arcadia_fonts_found() -> bool:
     return len(arcadia_fonts) > 0
 
 
-def save_figure(fname: str, context: str = "web", **savefig_kwargs: dict[Any, Any]) -> None:
-    "Save the current figure with the default settings for web."
+def save_figure(
```
keithchev, line 67: This is minor point but there's a bit of an inconsistency in the way this function is implemented, in that the validity of the filetype in `fname` is only checked if `fname_types` is provided. I would suggest slightly rewriting this function to first construct the `fname_types` list:
```py
suffix = Path(fname).suffix[1:]
if fname_types is None:
    if not suffix:
        raise ValueError("The filename must include a filetype if no filetypes are provided.")
    fname_types = {suffix}
else:
    fname_types = set(fname_types) | {suffix}
```
And then continue with the loop `for suffix in fname_types:`. This way more of the code and logic is the same, whether the user provides `fname_types` or not.

arcadia_pycolor/tests/test_mpl_save_figure.py
```diff
@@ -0,0 +1,59 @@
...
+
+@pytest.mark.parametrize(
+    "fname, fname_types, expected_outputs",
+    [
+        ("test.pdf", None, ["test.pdf"]),
+        ("test.pdf", ["pdf"], ["test.pdf"]),
+        ("test.pdf", ["pdf", "png"], ["test.pdf", "test.png"]),
+        ("test.pdf", ["png", "eps"], ["test.pdf", "test.png", "test.eps"]),
+        ("test.pdf", ["png"], ["test.pdf", "test.png"]),
+        ("test", None, ["test.pdf"]),
```
keithchev, line 20: This case is a bit surprising. I wonder if we should disallow this, even if matplotlib allows it (in other words, if `fname_types` is not provided, require that the filename includes a suffix).

keithchev (approved):
Looks good! Just one comment/thought about how to slightly improve how `save_figure` is implemented.

arcadia_pycolor/tests/test_mpl_save_figure.py
reillyowencooper: I agree, I think explicitly making the user put an extension if they don't provide a list definitely makes it more predictable.

arcadia_pycolor/tests/test_mpl_save_figure.py
```diff
@@ -0,0 +1,59 @@
...
+        ("test.invalid", ["pdf"]),
+    ],
+)
+def test_mpl_save_figure_fname_suffix_invalid(tmp_path, fname, fname_types, capsys):
+    simple_plot()
+
+    apc.mpl.save_figure(fname=tmp_path / fname, fname_types=fname_types)
+
+    captured = capsys.readouterr()
+    assert "Invalid file suffix 'invalid'. Skipping." in captured.out
```
reillyowencooper, line 52:
```suggestion
    assert "Invalid filetype 'invalid'. Skipping." in captured.out
```

reillyowencooper: (if following Keith's suggestion)

arcadia_pycolor/tests/test_mpl_save_figure.py
mezarque: Sounds good; will catch this case then instead of allowing it to pass through.

arcadia_pycolor/mpl.py
mezarque: Done! Thanks for sketching this out.

PR #53, opened 2024-08-14
Fix dev install command in README

lanery:
`env` needed in command to create a new environment. Otherwise results in
```
PackagesNotFoundError: The following packages are not available from current channels:

  - envs/dev.yml
```

## PR checklist


mezarque (approved):
Great catch! Thanks!

PR #56, opened 2024-10-21
Add to_plotly_colorscale

ekiefl:
I found this method useful so I decided to make this PR. I can write tests if the change is welcomed. I know you had been erring on the side of keeping this package lightweight, but in general, I think it would be nice to support some Plotly conversion so we could maintain our Arcadia look when using interactive plots.

### DIFF
--- a/arcadia_pycolor/gradient.py
+++ b/arcadia_pycolor/gradient.py
@@ -192,3 +192,22 @@ class Gradient(ColorSequence["Gradient"]):
             self.name,
             colors=colors,
         )
+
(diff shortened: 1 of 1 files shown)


keithchev (approved):
Looks fine by me! Even if it nominally broaches the plotly question, this is so lightweight and also does not introduce a dependency on plotly itself, so I think it's certainly harmless and will likely be useful for others.

mezarque:
Seems fine to me also! I'd love to be able to support Plotly in this package, but I don't have the bandwidth to extend it these days. If you wanted to take a stab, that would be awesome.

Issue #57, opened 2025-02-13
Calling `apc.mpl.setup` is slow (a few seconds)


keithchev:
This function only has to be called once per session/python process, but it's not ideal that it's so slow. We should figure out why it's slow and refactor for efficiency if possible.

PR #58, opened 2025-02-20
Speed up `load_fonts` on macOS

wasimsandhu:
## Summary

This PR restricts the font search to [the expected locations for installed fonts](https://support.apple.com/guide/font-book/change-font-book-settings-fntbk1004/mac#:~:text=All%20Users%3A%20Fonts%20are%20available,Library%20(%2FLibrary%2FFonts%2F).) on macOS systems:

- `/Library/Fonts`
- `~/Library/Fonts`
- `/System/Library/Fonts`
 
If no fonts are found in these directories, it falls back to the existing `matplotlib.font_manager._get_macos_fonts`  implementation.

This decreases the execution time for `apc.mpl.setup` from 11 sec to less than 1 sec on my machine.

Completes #57.

## TODO

- [x] Test if `load_fonts` is slow on Linux machines as well.

## PR checklist


mezarque:
Thanks for diagnosing this! I'm relieved that it wasn't just me doing something foolish (but instead is `matplotlib` doing something foolish) that caused the slowness of imports. Perhaps suggesting a fix to `matplotlib` could also be a good way to help them out?

wasimsandhu:
@mezarque That's not a bad idea!

wasimsandhu:
@keithchev I tested my changes on an Ubuntu 24.04 VM by running:

```python
import arcadia_pycolor as apc
apc.mpl.setup()
```

Fonts were loaded in less than 1 second, both on this branch and `main`, so it looks like this slowness is a bug with the macOS search specifically.

We are using an older version of `matplotlib` (3.9.1) but at a glance, it doesn't seem like the logic for searching font dirs has changed in matplotlib.font_manager (see [here](https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/font_manager.py#L265)).

arcadia_pycolor/mpl.py
```diff
@@ -416,10 +444,22 @@ def load_fonts(font_folder: Union[str, None] = None) -> None:
         font_folder (str, optional): the folder to search for fonts in.
             Uses the default system font folder if None.
     """
-    for fontpath in font_manager.findSystemFonts(fontpaths=font_folder, fontext="ttf"):
-        if FONT_FILTER.lower() in fontpath.lower():
-            font_manager.fontManager.addfont(fontpath)
-            font_manager.FontProperties(fname=fontpath)
+    arcadia_font_paths = []
+
+    # On macOS, look in the standard font locations first.
```
keithchev, line 449: The "first" here seems misleading, since on macOS, we *only* look in the standard locations.

arcadia_pycolor/mpl.py
```diff
@@ -416,10 +444,22 @@ def load_fonts(font_folder: Union[str, None] = None) -> None:
...
+    # This is faster than matplotlib's font search implementation.
+    # See https://github.com/Arcadia-Science/arcadia-pycolor/pull/58.
+    if font_folder is None and sys.platform == "darwin":
+        arcadia_font_paths.extend(_find_macos_arcadia_fonts())
+    # If no fonts are found, fallback to full system search.
+    else:
```
keithchev, line 455: Something seems wrong here - either this comment is wrong or the else-block is wrong, because as written, we get to the else-block if we're not on a Mac, not if no fonts are found. It feels to me like the comment should be correct and instead of the else block there should be a new if-block:

```suggestion
    # If no fonts are found, fallback to full system search.
    if not arcadia_font_paths:
```

keithchev (changes requested):
Nice! Glad this was an easy fix. Can confirm it also speeds up loading my (mac) laptop. I think there's just one potential issue in the `load_fonts` function to resolve.

arcadia_pycolor/mpl.py
wasimsandhu: Yup, I changed the code but forgot to update the comments. Was just about to fix this when your review came in haha.

mezarque:
For some context, I think the reason we're using a slightly older version of `matplotlib` is that there was a bug in `matplotlib` color registration when we developed this package. (I tried to find the relevant GitHub issue but no luck). It might be worth upgrading to a newer version of `matplotlib` at some point.

keithchev:
> For some context, I think the reason we're using a slightly older version of matplotlib is that there was a bug in matplotlib color registration when we developed this package.

Yeah, I remember this too, though I forget the details. In any case, I think the deeper reason we're on matplotlib 3.9 is that we want pycolor to support a reasonable set of older versions of matplotlib, since we can't control the environments in which people may need to use this package. It looks like 3.9 came out in mid 2024, so I would guess that we'd need to stay on 3.9 for another year or so.

keithchev (approved):
Approving with one comment: can we remove all of the notebook changes from this PR? They shouldn't haven't been substantively changed by this PR, so it will just pollute the commit history.

Issue #52, opened 2024-08-01
Expand docstrings


mezarque:
Currently, a lot of the docstrings are one-liners without a description of potential input parameters. We should do some cleanup and fill in that documentation.

PR #59, opened 2025-02-24
Improve docstrings, refactor variable names, lint markdown

wasimsandhu:
## Summary

Improve docstrings, rename variables, and lint markdown files. Completes #52.

### Docstrings
Docstrings attempt to adhere to [PEP 257](https://peps.python.org/pep-0257/) and the [Google style guide for Python](https://google.github.io/styleguide/pyguide.html#s3.8.1-comments-in-doc-strings).

There is some redundancy between docstrings for classes and constructor methods; while [both should exist](https://peps.python.org/pep-0257/#:~:text=The%20docstring%20for%20a%20class,documented%20by%20their%20own%20docstring.), there doesn't seem to be a convention for preventing redundancy between the two.

Some functions were missing return types, so I added those as well.

### Variable renaming
I tried to constrain my renaming to private functions and non-default parameters, which are unlikely to be used / specified by users. Most of the name changes are for trivial things, like renaming `axis` parameters to `axes`, or differentiating between `filename` and `filepath`.

Some of the name changes are for default parameters (e.g. `axes`, which can be `None` in many functions). If we don't want breaking changes in this PR, we can revert these.

### Markdown linting
The markdown linting was done using [markdownlint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint). If we want, we can add this as a recommended extension in `.vscode/extensions.json`, but I don't think it's necessary.

## PR checklist


mezarque:
”axis” -> “axes” seems like a sensible change, but as you noted would be breaking to previous versions. Perhaps we should increment the minor version?

arcadia_pycolor/utils.py
```diff
@@ -33,14 +35,16 @@ def interpolate_x_values(y_values: list[float], round_digits: int = 3) -> list[f
 
 
 def is_non_decreasing(values: NumericSequence) -> bool:
-    """Determine if the numbers in `values` are in strictly non-decreasing order.
+    """
+    Determine if the numbers in `values` are in strictly non-decreasing order.
```
ekiefl, line 39: Shouldn't the one-liner be on the same line as opening triple quotes? See below:

"""A one-line summary of the module or program, terminated by a period.

Leave one blank line.  The rest of this docstring should contain an
overall description of the module or program.  Optionally, it may also
contain a brief description of exported classes and functions and/or usage
examples.

Typical usage example:

  foo = ClassFoo()
  bar = foo.function_bar()
"""

wasimsandhu: You're right, this was my mistake!

keithchev:
> ”axis” -> “axes” seems like a sensible change, but as you noted would be breaking to previous versions. Perhaps we should increment the minor version?

Yeah, definitely a new minor version. Why don't we merge this, #60, and #61, and then publish a new version?

keithchev (approved):
Looks good! Thanks for doing this; it's great to have proper and consistent docstrings.

Issue #25, opened 2024-06-14
Should palettes be iterables?


mezarque:
Currently, we use palettes just to group discrete colors together. It might be more useful to allow them to be accessed by their indices, e.g. `apc.palettes.accent[0:2]` might return a `Palette` containing the first two colors. I suppose users can do this using `apc.palettes.accent.colors[0:2]`, but that feels a little cumbersome.

ekiefl:
First of all, nice work you guys! The documentation is extremely helpful and thorough, and the API feels nice, and **above all**, the plots and color options look very nice.

Speaking as 0.5.0's likely first user here. @mezarque, I noticed this question you made  based on the title "Should palettes be iterables?", because when using the API I intuitively expected `Palette` to have length and be iterable. After seeing the `colors` attribute being accessed in the quickstart guide, I realized that `Palette` isn't actually a color sequence, but it holds a color sequence. Easy enough, but a little confusing.

IMO, the `ColorSequence` abstract base class should implement length (`__len__`) and be iterable (`__iter__` and `__getitem__`). This would be more in line with the class name (Color**Sequence**), since a sequence is a lengthy, indexable iterable.

Here are the method implementations that turn `ColorSequence` into a true sequence and supports your desire to have slice-based indexing:

```python
    @overload
    def __getitem__(self, index: int) -> HexCode: ...
    
    @overload
    def __getitem__(self, index: slice) -> T: ...
    
    def __getitem__(self, index: Union[int, slice]) -> Union[HexCode, T]:
        if isinstance(index, slice):
            return self.__class__(name=f"{self.name}_slice", colors=self.colors[index])
        else:
            return self.colors[index]

    def __iter__(self):
        return iter(self.colors)

    def __len__(self):
        return len(self.colors)
```

The change in action:

```
[ins] In [2]: len(apc.palettes.primary)
Out[2]: 12

[nav] In [4]: for color in apc.palettes.primary: print(color)
#5088C5
#F28360
#3B9886
#F7B846
#7A77AB
#F898AE
#73B5E3
#FFB984
#97CD78
#C85152
#F5E4BE
#BABEE0

[ins] In [5]: apc.palettes.primary[2:5]
Out[5]:

   seaweed #3B9886
   canary  #F7B846
   aster   #7A77AB

[ins] In [12]: apc.palettes.primary.reverse().colors == apc.palettes.primary[::-1].colors
Out[12]: True

[...]

PR #60, opened 2025-02-24
Inheritance teardown

ekiefl:
This introduces the changes proposed in this issue: https://github.com/Arcadia-Science/arcadia-pycolor/issues/25. A more in depth justification of these changes is provided in that issue.

Note: These changes were implemented using Claude's 3.7 GitHub integration capabilities. Specifically, the web interface where you can upload specific files from your codebase.

Note: #61 was also merged into this branch, which removes all types as string references.

Everyone below this line is generated by Claude.

---

# Migration Guide: ColorSequence Refactoring

## Overview

We've refactored the `arcadia_pycolor` package to use composition over inheritance for color collections. This refactoring removes the abstract `ColorSequence` base class and makes `Palette` and `Gradient` standalone classes. This guide explains the changes and potential impacts on your code.

## Key Changes

1. **Removed the `ColorSequence` Abstract Base Class**
   - `Palette` and `Gradient` are now standalone classes
   - Both classes still provide the same core functionality

2. **Enhanced `Palette` to be a True Sequence**
   - Added `__len__`, `__iter__`, and `__getitem__` methods
   - Palette objects can now be used like Python lists

3. **Redesigned `Gradient` to use Composition**
   - Created a new `Anchor` class representing color-value pairs
   - `Gradient` now contains a list of `Anchor` objects
   - Added backward compatibility properties for `colors` and `values`
   - Added a new `num_anchors` property

### Code that Should Continue to Work

The following patterns should continue to work without modification:

```python
# Creating palettes and gradients
palette = Palette("my_palette", colors)
gradient = Gradient("my_gradient", colors, values)

# Accessing colors and values
for color in palette.colors:
    # ...

for value in gradient.values:
    # ...

# Converting to matplotlib colormaps
cmap = palette.to_mpl_cmap()
cmap = gradient.to_mpl_cmap()

# Other common operations
reversed_palette = palette.reverse()
reversed_gradient = gradient.reverse()
```

### New Capabilities

You can now use `Palette` objects as sequences:

```python
# Get length
length = len(palette)

# Iterate directly
for color in palette:
    # ...

# Access by index
first_color = palette[0]
last_color = palette[-1]

# Slice to create new palettes
subset = palette[1:3]
every_other = palette[::2]
reversed_palette = palette[::-1]  # Alternative to palette.reverse()
```

### Type Changes

If your code explicitly checks for `ColorSequence` instances, you'll need to update it:

```python
# Old code
if isinstance(obj, ColorSequence):
    # ...

# New code
if isinstance(obj, (Palette, Gradient)):
    # ...
```

### New `Gradient` Properties

`Gradient` objects now have these new properties:

```python
# Access the number of anchors
num_anchors = gradient.num_anchors

# Access anchors directly (new capability)
for anchor in gradient.anchors:
    color = anchor.color
    value = anchor.value
```

## Summary

This refactoring makes the API more intuitive and aligns the implementation with the conceptual model: palettes as color sequences and gradients as color sequence generators. The changes maintain backward compatibility for most common use cases while adding new sequence capabilities to palettes.


arcadia_pycolor/gradient.py
```diff
@@ -160,12 +192,9 @@ def __add__(self, other: "Gradient") -> "Gradient":
         """
         Return the sum of two gradients by concatenating their colors and values.
         """
-        new_colors = []
-        new_values = []
-
         # If the first gradient ends with the same color as the start of the second gradient,
         # drop the repeated color.
-        offset = 1 if self.colors[-1] == other.colors[0] else 0
+        offset = int(self.colors[-1] == other.colors[0])
```
mezarque, line 197: I think the previous expression was a little more explicit and readable, but they result in the same outcome. Which is better to bake?

arcadia_pycolor/gradient.py
```diff
@@ -176,18 +205,19 @@ def __add__(self, other: "Gradient") -> "Gradient":
         )
 
     def __repr__(self) -> str:
-        longest_name_length = self._get_longest_name_length()
+        longest_name_length = max(len(anchor.color.name) for anchor in self.anchors)
```
mezarque, line 208: If only to keep things parallel to the updated `palette` class, I think this should be implemented as a method.

arcadia_pycolor/gradient.py
```diff
@@ -176,18 +205,19 @@ def __add__(self, other: "Gradient") -> "Gradient":
...
-                for color, value in zip(self.colors, self.values)
+                f"{anchor.color.swatch(min_name_width=longest_name_length)} {anchor.value}"
+                for anchor in self.anchors
             ]
         )
 
     def to_mpl_cmap(self):
-        colors = [(value, color.hex_code) for value, color in zip(self.values, self.colors)]
+        """Convert the gradient to a Matplotlib colormap"""
+        colors = [(anchor.value, anchor.color.hex_code) for anchor in self.anchors]
```
mezarque, line 220: Should this re-structuring also be made into a method? I imagine it would be useful for passing to `plotly`.

arcadia_pycolor/gradient.py
```diff
@@ -39,9 +57,24 @@ def __init__(self, name: str, colors: list[Any], values: Union[list[Any], None]
...
+    def values(self) -> list[float]:
+        return [anchor.value for anchor in self.anchors]
+
+    @property
+    def num_anchors(self) -> int:
+        """Returns the number of anchors in the gradient"""
+        return len(self.anchors)
 
     @classmethod
     def from_dict(
```
mezarque, line 80: The current methods for creating a `Gradient` follow from the previous approach to structuring gradients by using two different lists (`HexCode`s and values). Should we perhaps create a method that allows users to go from a list of `Anchor`s, as well? Or should we avoid exposing users to the `Anchor` class?

mezarque (commented):
This looks reasonable to me overall, and it's quite cool that we can automatically generate these kinds of refactors using Claude. I also don't think this implementation breaks anything from the user-facing side of the API, which is nice.

I added a couple of small questions throughout, but I'd love to get @keithchev 's take on this refactor as well.

arcadia_pycolor/gradient.py
ekiefl: The former just reminds me of the very common and redundant idiom:

```
True if len(x) > 10 else False
```

arcadia_pycolor/gradient.py
mezarque: After thinking a little more, if we were to go this route, it would be nice to be able to add methods to create `Anchor`s on the fly without having to create the associated `HexCode` objects first.

But intuitively, I think it's not actually necessary to allow users to create `Gradient`s from `Anchor` objects because that would be clunky and cumbersome. Which makes me wonder whether `Anchor` objects are strictly needed, if they would never be exposed to users.

arcadia_pycolor/gradient.py
mezarque: It seems like the newer version is a common and more Pythonic idiom, so I'm fine if we go with it.

arcadia_pycolor/gradient.py
ekiefl: I imagine Anchor as an internal representation. We don't need it if you don't like it. But I added it because I like zipping properties together in light-weight dataclasses, sort of like NamedTuples, except good.

I think in general, it's important to separate how an object is constructed versus how they are represented internally.

arcadia_pycolor/gradient.py
ekiefl: In my opinion, no. This increases cognitive load (another name to remember), adds another method, another docstring, and introduces an abstraction that likely outweighs the benefit. After all, it's a simple list comprehension.

Now, if we wanted to remove Anchor in favor of more primitive types, I think this zipped list would serve as a great class attribute.

arcadia_pycolor/gradient.py
mezarque: I don't have a strong argument against this implementation, so I'm fine to go for it!

arcadia_pycolor/gradient.py
ekiefl: I think having specifically tailored `.to_plotly_x` methods could provide a lot more leverage.

arcadia_pycolor/gradient.py
```diff
@@ -39,9 +57,24 @@ def __init__(self, name: str, colors: list[Any], values: Union[list[Any], None]
                 raise ValueError("The first value must be 0 and the last value must be 1.")
             if len(colors) != len(values):
                 raise ValueError("The number of colors and values must be the same.")
-            self.values = values
+            anchor_values = values
         else:
-            self.values = distribute_values(len(self.colors))
+            anchor_values = distribute_values(len(colors))
+
+        self.anchors = [Anchor(color, value) for color, value in zip(colors, anchor_values)]
```
keithchev, line 64: surprised the linter doesn't catch this
```suggestion
        self.anchors = [Anchor(color, value) for color, value in zip(colors, anchor_values), strict=True)]
```

arcadia_pycolor/gradient.py
```diff
@@ -39,9 +57,24 @@ def __init__(self, name: str, colors: list[Any], values: Union[list[Any], None]
...
+
+    @property
+    def colors(self) -> list[HexCode]:
```
keithchev, line 67: This attribute feels like it should be called `anchor_colors`, since the user is likely to expect that `some_gradient.colors` will contain...the colors in the gradient, not the initial colors used to create the gradient. Although `colors` was previously a public attribute of `Gradient`, iirc there was no good reason for this except that it was a result of the shared `ColorSequence` class (which is one more argument against that idea, imo). So TLDR I think `Gradient.colors` should have been private to begin with, and although renaming it now nominally breaks backwards compatibility, I think it's probably worth it to eliminate this potential point of confusion.

arcadia_pycolor/gradient.py
```diff
@@ -39,9 +57,24 @@ def __init__(self, name: str, colors: list[Any], values: Union[list[Any], None]
...
+        return [anchor.color for anchor in self.anchors]
+
+    @property
+    def values(self) -> list[float]:
```
keithchev, line 71: Similarly this would become `anchor_values`

arcadia_pycolor/tests/test_palette.py
```diff
@@ -66,3 +66,57 @@ def test_palette_add():
...
+    sliced = palette[-2:]
+    assert len(sliced) == 2
+    assert sliced.colors == colors[-2:]
+
+    sliced = palette[::2]
+    assert len(sliced) == 2
+    assert sliced.colors == [colors[0], colors[2]]
+
+    palette = Palette("test", colors)
+    assert palette.reverse().colors == palette[::-1].colors
```
keithchev, line 122: I imagine the robot wrote this, but this is an elegant test.

arcadia_pycolor/gradient.py
keithchev: Yeah, I agree that `Anchor` is fine as an internal representation but shouldn't be user-facing.

keithchev (changes requested):
Looks good! The only caveat is that I don't fully remember why we used a shared base class for `Gradient` and `Palette` to begin with, but iirc there was no specific non-obvious reason, and I do agree that it now appears to be a forced and clunky abstraction.

One small final change is to update the docs. afaict, there are just a few references to the `colors` attribute of `Palette` in `quickstart.md` and `color_usage.ipynb` that should be replaced with direct indexing. It would also be good to formally document that `Palette` is a sequence - probably a brief mention in `color_usage.ipynb` would be sufficient.

arcadia_pycolor/gradient.py
keithchev: Edit: oops, nevermind. `strict` was added in 3.10 🤦‍♂️

arcadia_pycolor/gradient.py
ekiefl: It's a really low-lift method so I removed it in 21c5b82a35cbd10db97493184b4bde55c834fa8c. Rationale: 1. the method's only used once 2. its just a comprehension 3. it displaces the definition from the variable declaration.

ekiefl:
Ok I merged in #60, addressed comments, and matched our .pre-commit-config.yaml and Makefile to better match the python package template repo.

keithchev (approved):
Awesome! Thanks so much for doing this and for updating the docs so thoroughly. Having the palettes be true sequences is a really satisfying change.

Issue #38, opened 2024-07-03
Add notebook execution to CI


keithchev:
This is needed to verify that the documentation and example notebooks can be run without errors.

PR #63, opened 2025-02-27
Add notebook execution to CI

wasimsandhu:
## Summary

Execute notebooks in CI workflow. Completes #38.

## Changes

- Prefix `jupyter` command with `poetry run` in makefile.
- In CI environment, exit early if there are any notebook errors.
- Fetch entire commit history and tags in order to install the package in CI (see [here](https://github.com/mtkennerly/poetry-dynamic-versioning/discussions/132)).
- Run `make execute-all-notebooks` in CI.

## Tests

- Verified that the `make` command exits early on notebook execution error in CI environment. See [this job](https://github.com/Arcadia-Science/arcadia-pycolor/actions/runs/13638825344/job/38123805950).
- Re-ran the `execute-all-notebooks` command locally to verify that there were no regressions.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -35,15 +35,14 @@ jobs:
           virtualenvs-in-project: true
 
       - name: Cache Poetry dependencies
-        id: cache-poetry-deps
+        id: cache-deps
         uses: actions/cache@v2
         with:
           path: .venv
           key: pydeps-${{ hashFiles('**/poetry.lock') }}
 
-      - run: poetry install --no-interaction --no-root
-        if: steps.cache-poetry-deps.outputs.cache-hit != 'true'
-
+      - run: poetry install --no-interaction --no-root --with=dev
+        if: steps.cache-deps.outputs.cache-hit != 'true'
       - name: Run Ruff
         run: poetry run ruff check --output-format=github .
 
@@ -52,3 +51,14 @@ jobs:
 
       - name: Run tests
         run: poetry run pytest
+
+      # This is necessary to prevent the following error when installing the package in CI:
+      # "RuntimeError: This is a shallow repository, so Dunamai may not produce the correct version."
+      - name: Fetch full Git history
+        run: git fetch --unshallow --tags
+
+      - name: Install arcadia-pycolor
+        run: poetry run pip install .
+
+      - name: Execute notebooks
+        run: make execute-all-notebooks
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -27,7 +27,7 @@ jobs:
           python-version: "3.9"
 
       - name: Cache Poetry install
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: ~/.local
           key: poetry-1.8.3-0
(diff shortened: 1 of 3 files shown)


.github/workflows/lint-and-test.yml
```diff
@@ -52,3 +51,14 @@ jobs:
 
       - name: Run tests
         run: poetry run pytest
+
+      # This is necessary to prevent the following error when installing the package in CI:
+      # "RuntimeError: This is a shallow repository, so Dunamai may not produce the correct version."
```
keithchev, line 56: Could we change this to be a bit more descriptive? dunamai is a transitive dependency, so the origin of this error is a bit opaque. Something like this:
```suggestion
      # Fetching the full git history is necessary because poetry's dynamic versioning extension
      # requires it, and this extension is run even when the package is pip-installed locally (as we do below).
```

.github/workflows/lint-and-test.yml
```diff
@@ -52,3 +51,14 @@ jobs:
...
+      - name: Fetch full Git history
+        run: git fetch --unshallow --tags
+
+      - name: Install arcadia-pycolor
+        run: poetry run pip install .
+
+      - name: Execute notebooks
+        run: make execute-all-notebooks
```
keithchev, line 64: Do you know if this will exit with an error if one of the notebooks fail to execute? (I'm never sure how `make` might interfere with exit status propagation)

.github/workflows/lint-and-test.yml
```diff
@@ -52,3 +51,14 @@ jobs:
 
       - name: Run tests
         run: poetry run pytest
+
+      # This is necessary to prevent the following error when installing the package in CI:
+      # "RuntimeError: This is a shallow repository, so Dunamai may not produce the correct version."
+      - name: Fetch full Git history
+        run: git fetch --unshallow --tags
```
keithchev, line 58: To avoid this extra step, I think it's possible to specify a fetch depth in the initial checkout step:
```
      uses: actions/checkout@v3
      with:
        fetch-depth: 0
```

keithchev (approved):
Just a couple of suggestions for increased clarity/compactness.

.github/workflows/lint-and-test.yml
wasimsandhu: Oof, good catch. Even when a notebook has errors, this `make` command exits with status code 0 on my local machine. I fixed this by setting the command to exit early in CI if there are any errors:

```makefile
.PHONY: execute-all-notebooks
execute-all-notebooks:
	@if [ -n "$$CI" ]; then set -e; fi; \
	for file in $(JUPYTER_NOTEBOOKS); do \
		echo "Executing notebook $$file"; \
		poetry run jupyter execute --inplace $$file; \
	done
```

Tested in CI environment and it works. See [this job](https://github.com/Arcadia-Science/arcadia-pycolor/actions/runs/13638825344/job/38123805950).

wasimsandhu: Ideally, this command should finish executing all notebooks before exiting with a non-zero status code. Off the top of my head, we could keep a count of the failures and exit after the for loop if `fails > 0`. I'll open an issue to tackle this in a separate PR.

keithchev: I dunno, I think it's enough that the CI fails. A count of the failing notebooks alone won't be that informative, and outputting more than that could be complex. If we really care about this, we could probably use pytest to write a test that executes each notebook.

Issue #64, opened 2025-03-06
Add support for Plotly plots


wasimsandhu:
See [Theming and templates in Python](https://plotly.com/python/templates).

wasimsandhu:
Picking this up today.

Issue #54, opened 2024-08-22
Implement `__copy__` and `__deepcopy__` methods for `HexCode`


mezarque:
## Overview
When trying to use `HexCode` objects in `plotly` (not currently supported), we can run into problems caused by the way that `plotly` handles color objects; namely, that it uses `deepcopy` at some point in the figure generation process to create a copy of an input color object. Because of the way we developed `HexCode` to inherit from `str`, we don't currently override the `__copy__` or `__deepcopy__` methods. My guess is that this results in a problem when `deepcopy` is used on a `HexCode` object.

## Example
Example Plotly code that causes an error:
```python
import plotly.express as px

import arcadia_pycolor as apc

x = [1, 2, 3, 4, 5]
y = [1, 2, 3, 4, 5]
c = ['A', 'B', 'A', 'A', 'B']

cmap = {
    'A': apc.teal,
    'B': apc.tangerine,
}

px.scatter(x, y, color = c, color_discrete_map=cmap)
```

Full traceback of error:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[5], [line 15](vscode-notebook-cell:?execution_count=5&line=15)
      [8](vscode-notebook-cell:?execution_count=5&line=8) c = ['A', 'B', 'A', 'A', 'B']
     [10](vscode-notebook-cell:?execution_count=5&line=10) cmap = {
     [11](vscode-notebook-cell:?execution_count=5&line=11)     'A': apc.teal,
     [12](vscode-notebook-cell:?execution_count=5&line=12)     'B': apc.tangerine,
     [13](vscode-notebook-cell:?execution_count=5&line=13) }
---> [15](vscode-notebook-cell:?execution_count=5&line=15) px.scatter(x, y, color = c, color_discrete_map=cmap)

[...]

PR #67, opened 2025-03-26
Add support for Plotly plots

wasimsandhu:
## Summary

Add support for styling Plotly figures through the `apc.plotly` module. Completes #64.

## Changes

- Create a Plotly template layout in `style_defaults` that follows our style guide.
- Add a `plotly` module with high-level `setup`, `style_plot`, and `style_legend` methods that should handle most of the plot styling.
- Add methods for styling axes, ticks, and tick labels.
- Add methods for setting figure dimensions to the figure sizes described by  `full_wide`, `float_wide`, `full_square`, `float_square`, or `half_square` (e.g. 1000 px for `full_wide`).
- Add `save_figure` method for saving plots in various formats.
- Added tests and minimal documentation.

## Styles implemented

- Key title font: Suisse Int'l, semibold, 16 pt
- Axis titles font: Suisse Int'l, medium, 15 pt
- Body font: Suisse Int'l, regular, 15 pt
- Numbers font: Suisse Int'l Mono, regular, 14.5 pt
- Padding around plot: 40 px
  - Since the interactive plots won't have a white border around them, Audrey recommends adding an extra 20 px margin so that the margins match those of static plots (which have 20 px around the data and another 20 px around the panel).
- Margin between ticks and tick labels: 5 px
- Margin between tick labels and axis titles: 10 px

## Caveats

The biggest caveat with this PR is that **margins are not always applied correctly,** which in turn can result in **incorrect figure sizes**.

This can happen when tick labels are too long (like in `docs/examples/barplot.ipynb`), which causes the plot sizes to differ from the expected, predefined value. Instead of the defined margin being applied around the div, the margin is reduced as needed to result in the defined width and height.

**For plot exports,** the solution is to remove margins and adjust figure dimensions (in `apc.plotly.save_figure`). This ensures that the plot size (referred to as the "data panel" in the style guide) is correct. Audrey can then apply the margin in Illustrator herself. Later, we can implement a solution outside of Plotly to apply margins as expected (see #69).

**For interactive plot renders,** we can open an issue to investigate and better understand Plotly's automargin strategy, but this is likely a nontrivial task. For now, it will be the user's responsibility to adjust and refine margins as necessary. This can be done using `fig.update_layout`. For example:

```python
fig.update_layout(
    margin=dict(
        t=TOP_MARGIN,
        b=BOTTOM_MARGIN,
        l=LEFT_MARGIN,
        r=RIGHT_MARGIN,
    )
)
```

Additionally, the following items cannot be completed (as far as I can tell):

- Adding a line under the legend title.
- Setting the size of the trace symbol in the legend.
- Setting the margin between the trace symbol and text entries in the legend.
- Setting padding for the hovercard box.

I discussed these with @a-b-e-l-l and we agreed that it's okay to forgo these styles.

## Driveby Changes

- Completes #54.

## PR checklist


wasimsandhu:
@keithchev @mezarque @ekiefl

Requesting your review for this PR in its current state because I think it'd be good to get some eyes on the methods in `apc.plotly`, specifically whether everything a user would need to style their plot is available to them.

Don't let the large diff size fool you. The only two files that require review are `plotly.py` and `style_defaults.py`.

I listed some caveats and TODO's in the description. Feel free to leave your opinion on these as well; I plan to complete them async and re-request review from maybe just Keith when I'm done.

mezarque:
I think this seems fine overall - the functionality mirrors the core features of the matplotlib version and will save folks a lot of headaches. although I think it could be nice to address #54 in this so that you don't have to keep appending `.hex_code` to the end of `HexCode` objects in the example notebooks.

arcadia_pycolor/plotly.py
ekiefl, line 1: I feel it's bad practice to name modules after package dependencies, and can lead to annoyances like:

```python
(arcadia-pycolor) ~/Software/arcadia/arcadia-pycolor
▶▶ python /Users/evan/Software/arcadia/arcadia-pycolor/arcadia_pycolor/plotly.py
Traceback (most recent call last):
  File "/Users/evan/Software/arcadia/arcadia-pycolor/arcadia_pycolor/plotly.py", line 3, in <module>
    import plotly.graph_objects as go
  File "/Users/evan/Software/arcadia/arcadia-pycolor/arcadia_pycolor/plotly.py", line 3, in <module>
    import plotly.graph_objects as go
ModuleNotFoundError: No module named 'plotly.graph_objects'; 'plotly' is not a package
```

arcadia_pycolor/plotly.py
```diff
@@ -0,0 +1,471 @@
+from typing import Any, Literal, Union
+
+import plotly.graph_objects as go
+import plotly.io as pio
+from narwhals import Unknown
```
ekiefl, line 5: Why `narwhals`? It's meant for type safety for dataframe libraries.

arcadia_pycolor/plotly.py
```diff
@@ -0,0 +1,471 @@
...
+    capitalize_legend_title(fig)
+    capitalize_legend_entries(fig)
+
+
+def justify_legend_text(fig: go.Figure) -> None:
+    """Justify the legend to the left and change legend title font to Medium weight."""
+    fig.update_layout(legend_title_font_weight=600, legend_title_font_size=TITLE_FONT_SIZE)
+
+
+def get_styles(key: Union[str, None] = None) -> Union[dict[Unknown, Unknown], Unknown]:
```
ekiefl, line 365: Do `go.Layout` objects have accession methods? At first glance this seems like an API on top of an API 🤷.

I'd also favor against using a third party type in place of  what is effectively `Any`, but I'd be interested to hear a justification--I've never used `narwhals`.

ekiefl (approved):
The plots look great. Thanks for doing this. I left some food for thought.

arcadia_pycolor/plotly.py
```diff
@@ -0,0 +1,471 @@
...
+
+
+def capitalize_legend_title(fig: go.Figure) -> None:
+    """Capitalizes the legend title."""
+    fig.update_layout(legend_title_text=fig.layout.legend.title.text.capitalize())  # type: ignore
+
+
+def capitalize_legend_entries(fig: go.Figure) -> None:
+    """Capitalizes the legend entries."""
+    pass
```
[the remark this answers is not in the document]
wasimsandhu: Sorry, this was unfinished. Implemented now.

arcadia_pycolor/plotly.py
```diff
@@ -0,0 +1,471 @@
...
+
+def capitalize_legend_text(fig: go.Figure) -> None:
+    """Capitalize the legend title and entries."""
+    capitalize_legend_title(fig)
+    capitalize_legend_entries(fig)
+
+
+def justify_legend_text(fig: go.Figure) -> None:
+    """Justify the legend to the left and change legend title font to Medium weight."""
+    fig.update_layout(legend_title_font_weight=600, legend_title_font_size=TITLE_FONT_SIZE)
```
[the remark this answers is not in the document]
wasimsandhu: Also unfinished -- this styling is already set in the Plotly layout template, so this function was removed.

arcadia_pycolor/plotly.py
wasimsandhu: Whoops, that was a mistake. I imported it while writing the function and assumed it must be a type from the `typing` library.

arcadia_pycolor/plotly.py
wasimsandhu: Fair. You can access the attributes of the `go.Layout` method using dot notation, but they aren't typed. For example, the type hint for:

```py
legend = ARCADIA_PLOTLY_TEMPLATE_LAYOUT.legend
```

is:

```
(variable) legend: Unknown | tuple[Unknown, ...] | Layout | None
```

If you try to access `legend.title`, you'll get a linter error.

The purpose of this method was to provide the user a way to set styles themselves if the methods don't work for whatever reason. For example:

```py
go.Heatmap(
    colorbar=dict(
        **apc.plotly.get_colorbar_styles(),
        title_text="Intensity",
    ),
),
```

But I think you're right that implementing our own accession system is overkill. I'll replace this with:

```py
def get_arcadia_styles() -> dict[str, Any]:
    """Returns the Arcadia Plotly layout template as a dictionary."""
    return ARCADIA_PLOTLY_TEMPLATE_LAYOUT.to_plotly_json()
```

arcadia_pycolor/plotly.py
wasimsandhu: Ooh, good point. Let me think about what a better name would be. We could use `px` but this is a shorthand for a subset of the library, and not the library itself.

wasimsandhu: We discussed this offline, and per your suggestion, I renamed the module to `plotly_utils.py`. Then, in `__init__.py`, I rename the module on import as `plotly`:

```py
from arcadia_pycolor import plotly_utils as plotly
```

PR #70, opened 2025-03-29
Replace shorthand CSS font styles in SVG exports for Adobe Illustrator

### LINKED ISSUE #68: Plots saved as PDFs or SVGs do not display correct fonts in Adobe Illustrator
When plots exported with `apc.mpl.save_figure` are opened in Adobe Illustrator, axis titles are expected to have the Suisse Int'l font, but instead render a fallback font.

This issue was first reported by @sunandascript and @a-b-e-l-l in December 2024 in the `#software-questions` channel (link [here](https://arcadiascience.slack.com/archives/C04AKK76CDN/p1734729728400929)). The below screenshot displays an SVG in Illustrator (left) and in Apple Preview (right).

Note the incorrect font for the axis titles on the left.

Image

Strangely, the monospace font appears to be unaffected. It's not clear why, but we should investigate this and deliver a fix.

wasimsandhu:
## Summary

Replace shorthand CSS font styles (e.g. `500 15px SuisseIntl`) with individual CSS font properties (e.g. `font-family: SuisseIntl; font-size: 15px; font-weight: 500;`) to allow them to be rendered in Adobe Illustrator.

Fixes #68 for SVG exports.

## PR checklist

### DIFF
--- a/arcadia_pycolor/mpl.py
+++ b/arcadia_pycolor/mpl.py
@@ -101,11 +101,11 @@ def _find_macos_arcadia_fonts() -> list[str]:
 def _fix_svg_fonts_for_illustrator(filename: str) -> None:
     """Replaces shorthand font styles in SVG exports with individual CSS font properties.
 
-    This is needed because Adobe Illustrator does not respect font weights in shorthand font styles
-    like "font: 500 15px SuisseIntl, sans-serif;". The axis titles and legend title had some font
-    weight applied, and because of this, they were not being rendered in Illustrator.
+    This is needed because Adobe Illustrator cannot parse font weights in shorthand font styles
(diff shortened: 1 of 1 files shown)


wasimsandhu:
@keithchev I don't think we can provide a similar fix for PDF exports. Should we print a warning when the user tries to export to PDF with `apc.mpl.save_figure` instead?

keithchev:
> @keithchev I don't think we can provide a similar fix for PDF exports. Should we print a warning message when the user exports to PDF with apc.mpl.save_figure for now?

I'm not sure; I think it depends on whether folks already know to export to SVG for importing to Illustrator, in which case it might just be annoying. Maybe it's okay to wait and if this becomes a point of confusion we can add the warning?

PR #75, opened 2025-04-14
Set bounding box dimensions in `apc.mpl.save_figure`

wasimsandhu:
## Description

Dimensions for matplotlib plot exports are off by 6 or 7 pixels from the configured size. This appears to be because matplotlib is estimating the bounding box (bbox) dimensions itself. From the [`matplotlib.pyplot.savefig` docs](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html):

> #### bbox_inches str or [Bbox](https://matplotlib.org/stable/api/transformations.html#matplotlib.transforms.Bbox), default: [rcParams["savefig.bbox"]](https://matplotlib.org/stable/users/explain/customizing.html?highlight=savefig.bbox#matplotlibrc-sample) (default: None)
>> Bounding box in inches: only the given portion of the figure is saved. If 'tight', try to figure out the tight bbox of the figure.

This PR manually sets `bbox_inches` in the call to `plt.savefig` in `apc.mpl.save_figure`. This results in plots with dimensions off by 0.0006 pixels instead of 6. (I am not sure why.)

I exported a few plots with `apc.mpl.save_figure` and confirmed with @a-b-e-l-l that the dimensions are correct.

As a note, this will introduce breaking changes by requiring `figure_size` to be provided in the `apc.mpl.save_figure` call.

## PR checklist


wasimsandhu:
> How can the dimensions be off by a fraction of a pixel?

I had a hunch that there might not be enough decimal places in the `FIGURE_SIZES_IN_INCHES` values and I was right. Increasing them to 8 decimal places results in the exact correct dimensions in Illustrator, whereas 5 decimal places results in a fraction pixel offset.

PR #73, opened 2025-04-11
Apply styles to 3D Plotly figures

wasimsandhu:
## Summary

Extend methods in `apc.plotly` to work with 3D figures in Plotly, which use `go.layout.Scene` to apply styles.

Additionally, make some quality-of-life improvements to `apc.plotly.style_plot` based on feedback.

Completes #72.

Screenshot 2025-04-11 at 3 57 41 PM

## Changes

- Set default axis styles for 3D scenes in `style_defaults.py`.
  - @a-b-e-l-l and I agreed to show gridlines for 3D plots by default, since the plots may be harder to interpret without them.
- Extend all methods in `apc.plotly` to apply styles to 3D plots, if detected.
- Add 3D scatter plot example and plotting tests.

## Driveby Changes

- In the `capitalize_` methods, only capitalize if not already capitalized.
  - I noticed #55 is still open. We should make this change in `apc.mpl` as well before publishing a new release.
- In `style_plot`, capitalize all axis labels.
- In `style_plot`, add commas to ticklabels by default for monospaced axes.
  - This isn't done by default in `apc.mpl`, but seems like it should be. Thoughts?
- In `style_plot`, remove `colorbar_exists` argument and style automatically if detected.
- Add scatter plot example.

## PR checklist

- [x] Tag the issue(s) or milestones this PR fixes (e.g. `Fixes #123, Resolves #456`).
- [x] Describe the changes you've made.
- [x] Describe any tests you have conducted to confirm that your changes behave as expected.
- [x] If you've added new software dependencies, make sure that those dependencies are included in the appropriate `conda` environments.
- [x] If you've added new functionality, make sure that the documentation is updated accordingly.
- [x] If you encountered bugs or features that you won't address, but should be addressed eventually, create new issues for them.

Issue #55, opened 2024-09-06
Alter methods for capitalizing text


mezarque:
### Problem

Currently, we use the built-in `str.capitalize` method to capitalize text. This converts any string to the "Xxxxxx" capitalization style. For text that doesn't contain abbreviations or proper names, this is fine. But for text that contains capitalization (e.g. "PC1"), we end up with text that looks weird, e.g. "Pc1".

We should stop using `str.capitalize` and instead take an alternate route. For example, we could slice the first character of a string and set that to be capitalized, or detect whether any capitalization already exists within the string, and if so, avoid applying capitalization.

### Workaround

For now, users can manually run the sub-functions within `apc.mpl.style_plot`, rather than using the entire function (which includes capitalization).

For example, replace:
```python
apc.mpl.style_plot(axis = ax, monospaced_axes = "x", categorical_axes = "y")
```

with:

```python
apc.mpl.set_xticklabel_monospaced(axis = ax)
apc.mpl.set_yaxis_categorical(axis = ax)
```

PR #76, opened 2025-04-20
Only capitalize if no uppercase letters

wasimsandhu:
## Summary

For all capitalization methods in `apc.mpl`, only capitalize text when text has no uppercase letters. This fixes the issue of incorrectly changing text like "PC1" to "Pc1", or "mRNA" to "MRNA".

As a driveby change, performs the same change to the `capitalize_ticklabels` methods in `apc.plotly` (I appeared to have missed these in #73).

Addresses #55.

## PR checklist


keithchev (changes requested):
I'm not sure this is quite right because `isupper()` returns true only if *all* characters are uppercase, so this will fix the "PC1" -> "Pc1" problem but not the "mRNA" -> "MRNA" problem. I think it would be more robust to check if *any* letters in the label are capitals, and then only call `capitalize` if there are no capital letters in the label. FWIW, this is also one of the approaches suggested in #55.

wasimsandhu:
> I think it would be more robust to check if any letters in the label are capitals, and then only call capitalize if there are no capital letters in the label. FWIW, this is also one of the approaches suggested in https://github.com/Arcadia-Science/arcadia-pycolor/issues/55.

@keithchev Ah yeah, this is what I intended to do, but `isupper` was a mistake. Changed all instances of `not {text}.isupper()` to `{text}.islower()` to check if all characters in a string are lowercased. Also updated docstrings. Appears to be working as intended with both examples I offered.

keithchev (approved):
Just a side note that at some point it might be worth de-duplicating this capitalization logic, simple as it is, since it is currently duplicated in 19(!) different places.

PR #85, opened 2025-08-26
Clarify size argument in save_figure() docstring

lanery:
Helps clarify the docstring for the size argument in `apc.mpl.save_figure()` by making it the same as in `apc.mpl.get_figure_dimensions()` where it's nice and clear:
```
size (FigureSize): The size of the figure, which must be one of the following:
    - "full_wide"
    - "full_square"
    - "float_wide"
    - "float_square"
    - "half_square"
```


lanery:
No problem. Is the lint check supposed to fail?

PR #107, opened 2026-06-22
Fix diverging gradients crashing matplotlib colormap registration

John-king-zhou:
## PR checklist

- [ ] Tag the issue(s) or milestones this PR fixes (e.g. `Fixes #123, Resolves #456`).
- [ ] If you've added new software dependencies, make sure that those dependencies are included in the appropriate `conda` environments.
- [ ] If you've added new functionality, make sure that the documentation is updated accordingly.
- [ ] If you encountered bugs or features that you won't address, but should be addressed eventually, create new issues for them.

## Description

`rescale_and_concatenate_values` rescales the left half of a diverging gradient to `[0, 0.5]` and the right half to `[0.5, 1]`. This means the last anchor of the left half and the first anchor of the right half both land at exactly `0.5`, producing a duplicate position value.

`matplotlib.colors.LinearSegmentedColormap.from_list` requires **strictly increasing** position values and raises a `ValueError` on any duplicate:

```
ValueError: the values passed in the (value, color) pairs must increase monotonically from 0 to 1.
```

This affects all three diverging gradients (`orange_sage`, `red_blue`, `purple_green`) and — as of v0.8.0 where `to_plotly_colorscale()` is called at module level in `style_defaults.py` — causes `import arcadia_pycolor` to fail entirely.

## Fix

In `Gradient.to_mpl_cmap`, nudge any duplicate anchor value forward by `epsilon = 1e-10` before passing the `(position, color)` pairs to `from_list`. The adjustment is visually imperceptible but satisfies matplotlib's strict-monotonicity requirement.

## Testing

Confirmed that `import arcadia_pycolor as apc` and `apc.mpl.setup()` both succeed after the fix, and that the diverging colormaps render correctly.

### DIFF
--- a/arcadia_pycolor/gradient.py
+++ b/arcadia_pycolor/gradient.py
@@ -253,7 +253,8 @@ class Gradient:
                 v = adjusted_values[-1] + epsilon
             adjusted_values.append(v)
         colors = [
-            (v, anchor.color.hex_code) for v, anchor in zip(adjusted_values, self.anchors, strict=True)
+            (v, anchor.color.hex_code) for v, anchor in zip(adjusted_values, self.anchors,
+                                                            strict=True)
         ]
         return mcolors.LinearSegmentedColormap.from_list(
             self.name,
--- a/arcadia_pycolor/gradient.py
+++ b/arcadia_pycolor/gradient.py
@@ -253,7 +253,7 @@ class Gradient:
                 v = adjusted_values[-1] + epsilon
             adjusted_values.append(v)
         colors = [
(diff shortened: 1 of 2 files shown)


wasimxyz (changes requested):
Could you add a regression test before we merge? The current diff only touches `gradient.py`, and nothing guards against this recurring. Something simple like importing the package and running `apc.mpl.setup()`, plus a test that builds each diverging gradient's colormap (`to_mpl_cmap()`), would lock it in.

wasimxyz (approved):

PR #113, opened 2026-07-31
Wire runtime __version__ to hatch-vcs

lanery:
Fixes an issue with `__version__`. Old behavior on a fresh installation:

```python
>>> import arcadia_pycolor as apc
>>> apc.__version__
'0.0.0'
```

## Summary
- Generate `arcadia_pycolor/_version.py` at build time via hatch-vcs so runtime `__version__` matches git tags
- Import the generated version in `__init__.py`, with a fallback to `"0.0.0"` when the file is missing
- Update `.gitignore` comment to reference hatch-vcs instead of setuptools_scm

## Test plan
- [x] `uv run python -c "import arcadia_pycolor; print(arcadia_pycolor.__version__)"` returns a git-derived version (not `0.0.0`)
- [x] `uv build` produces a wheel containing `arcadia_pycolor/_version.py` with the expected version string

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,7 +77,7 @@ venv/
 # OS
 .DS_Store
 
-# written by setuptools_scm
+# written by hatch-vcs
 **/_version.py
 
 # Nextflow related folders and files
--- a/arcadia_pycolor/__init__.py
+++ b/arcadia_pycolor/__init__.py
@@ -6,8 +6,10 @@ from .gradient import Gradient
 from .hexcode import HexCode
 from .palette import Palette
 
-# This is a placeholder that will be replaced by the version number at build time.
-__version__ = "0.0.0"
+try:
+    from ._version import __version__
+except ImportError:
+    __version__ = "0.0.0"
 
 __all__ = [
     "cvd",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,6 +49,9 @@ dev = [
 [tool.hatch.version]
 source = "vcs"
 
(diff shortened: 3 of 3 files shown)


wasimxyz (approved):
Looks good. Review from Grok 4.5:

> `0.0.0` was intentional as a **placeholder**, not as the real published version. Keeping it only as a **fallback** (as this PR does) is the right shape.
> 
> **Why it was `0.0.0`**
> - Added in `c37e91b` (“add dynamic versioning”) with poetry-dynamic-versioning.
> - Same idea as `version = "0.0.0"` in `pyproject.toml`: a dummy that build tooling was supposed to replace from git tags.
> - The comment said it would be replaced at build time; that only really applied to **package metadata**. Runtime `apc.__version__` never got wired up — and after the uv/hatch migration (#102), hatch-vcs only set the metadata version unless you add a `version-file` hook.
> 
> **Should it stay this way?**
> - **No** as the always-on runtime value — that’s the bug this PR fixes (`'0.0.0'` on a normal install).
> - **Yes** as the `ImportError` fallback when `_version.py` isn’t generated (bare source tree, no build hook). That matches the README: `0.0.0` means version inference failed.
> 
> **PR take:** Correct fix. Generating `arcadia_pycolor/_version.py` via hatch-vcs matches the existing `.gitignore` (`**/_version.py`) that was already waiting for this. Approving the approach; the remaining `0.0.0` is only the safety net, not the intended version.

